PR #9, opened 2015-11-14
[qt4] Provide a default filename when saving

knowledgejunkie:
Provide a default (selected) filename when saving video, or when saving
a video frame to BMP/JPEG.

The default filename is constructed from the basename of the current open
video, with an extension determined by the output container/image format.

If the user does not want to use the default filename, they can simply
start typing to replace the selected filename with one of their choosing.

This update also has the effect of filtering the files displayed in the
FileDialog to those having the same extension.


mean00:
Thanks

PR #10, opened 2015-11-14
[qt4] Avoid overwriting output files when saving video/images

knowledgejunkie:
Always prompt the user before overwriting a file in the output directory
which has the same filename, when file filtering in the FileDialog is enabled.

This change should be applied after the update to provide a default filename
(including extension) when saving video/images.

Currently:

When file filtering is enabled, a file extension is automatically
added when the output filename does not contain a dot ("."). This is also
the time when ADM checks if the output file exists after adding the
filtering extension.

With ADM updated to suggest a default filename when saving, this file check
might not happen (because the output filename will probably already contain
a dot (".")).

This update tries to ensure that the user is always prompted before
overwriting a file when saving.


mean00:
thx

PR #12, opened 2016-05-03
Fix typos and improve consistency of some error messages

knowledgejunkie:
Fix typos and improve consistency of some error messages.

I have tried to avoid changing translatable strings where possible.


mean00:
thank you

PR #13, opened 2016-07-03
Bandaid for narrowing error

ajschult:
http://avidemux.org/smif/index.php/topic,16900.0.html

this doesn't seem right, but shuold yield what older gcc was already doing.  cpu_caps does not seem to be used.


mean00:
Let me have a closer look at that one
It should not be necessary to do it

ajschult:
Obsoleted by
https://github.com/mean00/avidemux2/commit/b54303bc9b9310e288a1dde9c923f48e854bd55f

PR #16, opened 2016-07-24
Call ADM_vd6_cleanup on shutdown

ajschult:
There's a cleanup method for video decoders.  This calls the method during shutdown and fixes some additional leaks.


mean00:
thx

PR #18, opened 2016-09-24
Avoid crashing seeking for deleted keyframe

ajschult:
see
http://avidemux.org/smif/index.php/topic,11557.msg60516.html#msg60516
http://avidemux.org/smif/index.php/topic,17173.0.html


mean00:
thanks

PR #28, opened 2016-10-30
[x264/Qt][x265/Qt] Add translatable default none tuning

eumagga0x2a:
This patch adds null/none tuning as the default tuning for x264 and x265 encoders if the advanced configuration is disabled and makes the label "none" translatable. Furthermore, it adds `grain` to the `ListOfTunings` array for x265.

This is basically [[x265/Qt] Add none preset (eumagga0x2a)](https://github.com/mean00/avidemux2/commit/6bb59711f290291c532bd60f37513e3b04d763e3), but finally done right (hopefully).


mean00:
the "" part is a special case,
it would deserve a bit of comment, else next time someone will do a cleanup it will wreck havoc

eumagga0x2a:
I commented the place in the code performing the substitution. Should I have commented the presence of `""` where `listOfTunings` is defined as well?

mean00:
Great, that way it will not be "optimized"

PR #35, opened 2016-11-12
admDxva2::init() should fail when its call to IDirectXVideoDecoderSer…

mm1044:
…vice_GetDecoderDeviceGuids() fails.


mm1044:
(Untested...)

mm1044:
Or, if "return true" is actually wanted, we should document why and create a "goto succeedInit"?

PR #36, opened 2016-11-14
[Dxva2] lookupCodec(): after an unsuccessful mode configuration check…

mm1044:
…, continue checking next modes.


mm1044:
(Untested...)

mean00:
I'm not sure i'm getting the patch
What's the thing that is better with it ?

mm1044:
For example, (as I remember on my Win.Vista)
ADM supports 3 GUIDs for H264,
assume 1st doesn't match, 2nd matchs but has no supported config, then 3rd should be tried too.

mean00:
ok, i'll look into mergint it
Thanks

mean00:
I"ve done it slightly differently
See latest commit

PR #37, opened 2016-11-17
Remove (confusing) accel key from TimeShift left panel checkbox.

mm1044:
See [French translation](http://avidemux.org/smif/index.php/topic,17199.msg77460.html#msg77460)

### DIFF
--- a/avidemux/qt4/ADM_userInterfaces/ADM_gui/gui2.ui
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_gui/gui2.ui
@@ -703,7 +703,7 @@
              <item>
               <widget class="QCheckBox" name="checkBox_TimeShift">
                <property name="text">
-                <string>&amp;Shift:</string>
+                <string>Shift:</string>
                </property>
               </widget>
              </item>


eumagga0x2a:
In which way is this accel key confusing? Are you going to re-add the accel key via avidemux_en.ts? Otherwise I'd object to removing it before some reasonable keyboard accessibility has been implemented (currently the navigation toolbar intercepts a lot of important keys even if no video is loaded, which renders Avidemux GUI almost completely inaccessible via keyboard). Accel key conflicts should be resolved, of course. There is none in the English locale in this case.

mm1044:
I explained the misbehavior/confusion in the forum.
Would re-adding it to avidemux_en.ts improve its behavior?
(I would think that this accel key can be restored later, when it works fully right.)

mm1044:
(Untested...)

eumagga0x2a:
I can't spot any misbehaviour related to the accel key for the TimeShift chechbox with the English locale, especially I can't reproduce the mentioned interaction with the "File" menu, at least on Linux and on Windows 7.

Defining accel keys in translation files makes the source better searchable and the whole thing more flexible. A drawback would be a lack of accelerators if translations are not installed (e.g. on Linux they are often split into a separate package).

If there is an accel key conflict in the French locale, it should be resolved in avidemux_fr.ts.

Finally, your change means a string change and would require all .ts files to be regenerated (I don't know whether lupdate is clever enough to keep the existing translations, in worst case .ts files which have translations for `&amp;Shift:` would require manual editing).

PR #40, opened 2016-11-18
[Qt] Make the time display 4px wider to avoid clipping with breeze Qt theme

eumagga0x2a:
This should be enough to avoid clipping with the breeze theme.


mean00:
Looks ok, thanks
I'll do the release tomorrow, i wont have much time in the upcoming weeks

eumagga0x2a:
Thank you for catching this issue. Looking forward to the release.

PR #49, opened 2016-12-01
[x265] Use utf8 for filenames on Windows too

eumagga0x2a:
This fixes 2-pass encoding in HEVC failing on Windows when the specified filename contains diacritics or other non-ASCII characters. The patch is exactly the same solution as used in the x264 encoder.

Reference: [x265 pass 2 failure](http://avidemux.org/smif/index.php/topic,17296.0.html)

### DIFF
--- a/avidemux_plugins/ADM_videoEncoder/x265/CMakeLists.txt
+++ b/avidemux_plugins/ADM_videoEncoder/x265/CMakeLists.txt
@@ -1,7 +1,7 @@
 
 INCLUDE(ve_plugin)
 INCLUDE(ve_settings_plugin)
-
+ADD_DEFINITIONS(-DX265_USE_UTF8=1)
 SET(x265_SRCS 
         ADM_x265.cpp
         ADM_x265Plugin.cpp


mean00:
The problem is that the matching win32/win64 libx265 binary must support utf8
Which is not the case as far as i can see it
the version used at the moment is too old

eumagga0x2a:
I use v2.1 of libx265 in my MXE setup on Debian 8, what version are you using in yours? If it is older, it should not be a problem to update the lib, should it? In my brief testing of an 32 bit Avidemux build containing this changeset on Windows 7, I did not encounter any issues saving h265 in 2-pass mode to destinations with filenames containing at least diacritics (should have tested at least with some 漢字 or at least 한글 characters, I know).

mean00:
My point was unclear
1- Either the fact that x265 is supporting utf8 is checked and the USE_UTF8 flag is set accordingly
2- Either a more recent version is assumed, and the flag disappears altogether and the #ifdef is removed

Yes updating is not a problem, just takes a bit of time

eumagga0x2a:
If possible, I'd prefer the second option, especially if it would allow to remove the #ifdef alongside the `utf8StringToAnsi` function. No problem if it takes time. I'll close this pull request then.

PR #51, opened 2016-12-03
[core/prefs] Accept too long or incomplete partial parameter lists

eumagga0x2a:
Ignoring the return value of ADM_paramValidatePartialList prevents Avidemux from unnecessarily discarding configuration on version update. According to my testing with added or removed entries in config3 this is reasonably safe (the only crash I experienced – unrecoverable without fixing or deleting the config3 file manually  – happens independently of this patch). Type validation is provided to some degree within the `ADM_paramLoadInternal` function.

### DIFF
--- a/avidemux_core/ADM_coreUtils/src/ADM_paramList.cpp
+++ b/avidemux_core/ADM_coreUtils/src/ADM_paramList.cpp
@@ -258,7 +258,7 @@ bool ADM_paramValidatePartialList(CONFcouple *couples, const    ADM_paramList *p
     int p=XgetParamSize(params);
     if(n>p)
     {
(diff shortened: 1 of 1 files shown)


eumagga0x2a:
The crash mentioned above happens in JSON code if the line

`         "reset_encoder_on_video_load" : false`

in `~/.avidemux6/config3` is deleted, but the comma after the closing curly bracket at the line just before that not:

`Program received signal SIGSEGV, Segmentation fault.
0x00007ffff4f1a956 in JSONNode::name[abi:cxx11]() const (this=0xffffffff)
    at /home/eumagga/adm/git/avidemux2/avidemux_core/ADM_coreUtils/src/Source/JSONNode.h:641
641	    return internal -> name();`

PR #52, opened 2016-12-13
Cross-compiling with MXE (WIP)

eumagga0x2a:
The following patches add a bash script based on the current `bootStrapCrossMingw_w32Qt5_mxe.sh` and `bootStrap.bash` to automate cross-compiling of Avidemux for Windows and to collect all required libs together with the fontconfig configuration (unsure if the latter is needed, but official nightlies include it). They add also a brief how-to describing the setup of the MXE cross-compile environment.

The script adds many options from `bootStrap.bash` like partial builds and doesn't use any scripts from the `foreignBuilds` directory. Creation of debug builds is not yet implemented, I was unsure whether this functionality would be useful. Some libs and binaries like twolame, aften, vapoursynth and avsproxy are not yet available in my MXE setup.

Cross-compiling using the default MXE qt5 package requires a tweak to some include paths in `cmake/admCheckQt5.cmake`, the first of the patches introduces a `TYPE2_CROSS_SETUP` cmake env variable to avoid breaking your own setup.

Due to the build of the bundled libass failing with mingw (it looks like [[Plugin] Fix build with rasterizer file. The fix is bad, the difference is not linked to clang but to a difference on the build node. it will do for now](https://github.com/mean00/avidemux2/commit/2651a3d2f9d823e44ff607884475a13df9bf9e35) doesn't work in MXE), I would like to upstream a modified rpmfusion.org patch to allow linking against system libass, liba52 and libmad as an option. The patch doesn't change any defaults; Avidemux still builds fine with internal libs.


mean00:
Looks good

PR #56, opened 2016-12-17
[build/cmake] Fix NVENC detection and ffmpeg configure options

eumagga0x2a:
This fixes NVENC detection and ffmpeg configuration during a native build on Fedora. Tested working in MXE as well.


eumagga0x2a:
Oops, the status message in the first commit was misplaced.

PR #58, opened 2016-12-18
[UI] Remove forgotten unused variable and simplify handling of pts==0 in A_jumpToTime

eumagga0x2a:
Oops, there were some editing leftovers in the last commit. Also don't try to call `goToTimeVideo` with `pts` equal zero, it spits a ton of terrifying errors if it fails, just call `rewind()`.

My apologies for the trouble.


eumagga0x2a:
Correct a premature variable declaration too.

PR #68, opened 2017-01-09
[plugins/changeFps] Fix navigation with the slider in the filter preview window

eumagga0x2a:
The reimplemented `goToTime` function was pulling the rug from under `ADM_flyDialog::sliderChanged` resulting in a [confusing behaviour](http://avidemux.org/smif/index.php/topic,17393.msg78917.html#msg78917) of the navigation slider in the filter preview window. What was the primary reason to rescale time?

[[changeFps] Working filter](https://github.com/mean00/avidemux2/commit/037bd7ccdcd2e36a13c5aba7f09ebf09f1c8f0a8)

It seems to work fine without that.

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/changeFps/changeFps.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/changeFps/changeFps.cpp
@@ -59,7 +59,6 @@ protected:
 public:
                             changeFps(ADM_coreVideoFilter *previous,CONFcouple *conf);
                             ~changeFps();
-        bool                goToTime(uint64_t usSeek);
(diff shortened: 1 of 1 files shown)


mean00:
Fixed slightly differently

eumagga0x2a:
Thank you, the point of the new, correct fix is that it will work even after another timing-modifying filter, right?

I wonder if anything can be done with the wrong time display / slider scaling in the main window with changeFps when "play filtered" is active (either the total duration should reflect the effect of changeFps or the slider / time display should advance slower or faster to match settings in changeFps). This is not a new issue though.

PR #70, opened 2017-01-18
Greek translation 001

nikoss:
Hi, this is the Greek translation.
I did a general check with avidemux-qt-2.6.18-2.mga6.i586.rpm ,
and I didn't see any issues.

Also, you may want to check the bugs in sourceforge.net , where I have reported a few typos.


nikoss:
Oops, wrong file.
I'll try to reject the pull request

PR #71, opened 2017-01-18
Nikoss greek translation 001

nikoss:
Hi, this is the Greek translation.
This file has some differences compared to the previous "el.ts" file : special characters and tags(?).
But when creating the "el.qm" file, the above don't seem to be an issue.
I did a general check with avidemux-qt-2.6.18-2.mga6.i586.rpm ,
and I didn't find any problems.

---
Also, you may want to check the bugs in sourceforge.net , where I have reported a few typos.


mean00:
Thank you

PR #78, opened 2017-01-24
Making plugin strings translatable

TotalCaesar659:


eumagga0x2a:
Please see [http://doc.qt.io/qt-5/qtglobal.html#QT_TRANSLATE_NOOP](http://doc.qt.io/qt-5/qtglobal.html#QT_TRANSLATE_NOOP) for the correct usage of the `QT_TRANSLATE_NOOP` macro. I really recommend setting up a build environment for Avidemux, with your patch the source wouldn't even compile.

TotalCaesar659:
Sorry, it's just a rocket science for me. I did it after trying to translate this strings, but without a result. It should be better if you or @mean00 make it instead of me.

avidemux_plugins/ADM_videoFilters6/asciiView/ADM_vidAscii.cpp
```diff
@@ -56,8 +56,8 @@ DECLARE_VIDEO_FILTER(   AsciiFilter,   // Class
                         ADM_UI_ALL,         // UI
                         VF_TRANSFORM,            // Category
                         "AsciiView",            // internal name (must be uniq!)
-                        "Ascii View",            // Display name
-                        "Ascii view" // Description
+                        QT_TRANSLATE_NOOP("Ascii View"),            // Display name
```
eumagga0x2a: You must supply a context, e.g.

```QT_TRANSLATE_NOOP("asciiView","Ascii View"), // Display name```

In this case "asciiView" is an arbitrary string serving as context for the translation (identical strings may be translated differently if they have different context).

File: avidemux_plugins/ADM_videoFilters6/asciiView/ADM_vidAscii.cpp [hunk 2]
```diff
@@ -56,8 +56,8 @@ DECLARE_VIDEO_FILTER(   AsciiFilter,   // Class
...
+                        QT_TRANSLATE_NOOP("Ascii view") // Description
```
eumagga0x2a: Again, the context is missing (should be the same as above):

`QT_TRANSLATE_NOOP("asciiView","Ascii view") // Description`

TotalCaesar659:
You're totally right. I tried to do it like here https://github.com/mean00/avidemux2/blob/master/avidemux_plugins/ADM_videoFilters6/resampleFps/ADM_vidResampleFPS.cpp
This context was already there, I just forgot about it. Shame for me.
I'll create a new PR and close this one after.

PR #84, opened 2017-01-30
Updated Russian translation

TotalCaesar659:

PR #86, opened 2017-02-01
[i18n/ru] Correct a few translation, add kbd accels, remove a lot of broken lrelease heuristic leftovers

eumagga0x2a:
This patch addresses a few issues mentioned in [https://github.com/mean00/avidemux2/pull/84](https://github.com/mean00/avidemux2/pull/84) (mainly correcting the usage of the word "граница" as translation for "border" as agreed with the translator), removes a ton of missed leftovers caused by the broken lrelease heuristic (they all originate from the time prior to the [fix](https://github.com/mean00/avidemux2/commit/53f744224a843f12c8ecaabdeaa1b9f494de97d6)) and adds a few keyboard accelerators.

Please note that adding the "hue" filter crashes Avidemux without a usable backtrace when `LANG` is set to `ru_RU.UTF-8` at least on Linux. This crash doesn't happen e.g. with the English or German locale. Deleting `avidemux_ru.qm` helps. The first bad commit seems to be [Updated Russian translation](https://github.com/mean00/avidemux2/commit/497cbc52c3aa35621d7f1b55bcca8c209271d92f). I haven't found the direct cause yet, so this patch doesn't address the crash.

### DIFF
--- a/avidemux/qt4/i18n/avidemux_ru.ts
+++ b/avidemux/qt4/i18n/avidemux_ru.ts
@@ -4501,7 +4501,7 @@ Drop timing informations ?</source>
         <location line="+3"/>
         <location filename="../../../avidemux_plugins/ADM_videoFilters6/addBorder/ADM_vidAddBorder.h" line="+42"/>
         <source>Add Borders</source>
-        <translation type="unfinished">Добавление границ</translation>
+        <translation type="unfinished">Добавление полей</translation>
     </message>
     <message>
         <location line="+4"/>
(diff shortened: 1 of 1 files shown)


eumagga0x2a:
The translations triggering the crash are identified and removed for now, but the real bug could reside either in `ADMImage::printString` or somewhere below (an issue with non-latin characters?).

mean00:
Probably.
The printString assumes pure ascii as input, it may go out of bounds with non ascii chars

eumagga0x2a:
If there are no immediate plans to rewrite `printString` to support utf8 and maybe use freetype as rasterizer, I'd revert the relevant parts of [[i18n] Make more strings translatable (Daniel Amm)](https://github.com/mean00/avidemux2/commit/3972605aa5777d95afc5fce035fc47bfb030b6a2). It is pure luck that the German translation didn't use any umlauts here. Inviting contributors to translate strings which may not include any non-ASCII chars without telling them about that calls for trouble.

PR #93, opened 2017-02-13
[desktop integration] Add AppData file

adrienverge:
AppData is a standard to describe GUI applications. In particular, it allows these applications to be searched for and installed through software centers on Fedora, Ubuntu, etc.

This commit embeds a `.appdata.xml` file for Avidemux, so my grandmother can install it without having to open a terminal.

For further information on AppData see:
https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps

Related to: https://github.com/rpmfusion/avidemux/pull/6


hobbes1069:
Does this mean the GTK interface is supported again?

mean00:
Not really, but patches are welcome

PR #94, opened 2017-02-14
[delogo] Fix interaction between the rubber band, the spinners and the zoom (quick and dirty style)

eumagga0x2a:
This patch fixes in a quick and dirty way the current odd interaction between the spinners and the rubber band. It also reimplements `autoZoom` so that toggling autozoom doesn't change values anymore.


eumagga0x2a:
Thank you very much for merging the PR and [fixing](https://github.com/mean00/avidemux2/commit/7b941d19d9f0ec49db67990c187f13d6e72cd75a) the clumsiest lapse in the patch.

PR #104, opened 2017-02-26
[encoding] Use encoding dialog for the first pass (no re-use yet)

eumagga0x2a:
This patch falls way short of the goal to use the single instance of the encoding dialog for both the first and the second pass. It just replaces the working dialog with the encoding one for the first pass with related additional tweaks like disabling the "Advanced" tab and making information about remaining time available. The benefit: it is now possible to minimize Avidemux to the notification area during the first pass too (see [reference](http://avidemux.org/smif/index.php/topic,12493.0.html)). Without re-use of the dialog instance, the application window will pop up after the first pass though.

The patch makes also the "Pass 1" translatable and hides the disfunctional shutdown checkbox and priority  combo – on Linux and Mac completely, on Windows only for the first pass. The name of the priority label is changed to clarify its purpose when used in code elsewhere.

Automatic shutdown after  encoding can't work now, of course, as the value of `shutdownRequired` is not evaluated in the destructor and `ADM_shutdown` is never called. IMVHO, this functionality should be either fixed or the checkbox removed / hidden to avoid user confusion.

### DIFF
--- a/avidemux/common/gui_savenew.cpp
+++ b/avidemux/common/gui_savenew.cpp
@@ -73,6 +73,7 @@ protected:
         bool                  setupAudio();
         ADM_audioStream      *audioAccess[ADM_MAX_AUDIO_STREAM]; // audio tracks to feed to the muxer
         int                   nbAudioTracks;
+        Clock                ticktock;
         
 public:
                               admSaver(const char *out);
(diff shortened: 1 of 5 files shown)


mean00:
No problem, it was still on the todo

PR #105, opened 2017-03-04
[crashdump] Add missing newline after file name

eumagga0x2a:
I crashed Avidemux too often to further ignore this merely cosmetical omission in the output of `ADM_backTrack` for Linux and MinGW (the Apple version of the function has its trailing newline).

### DIFF
--- a/avidemux_core/ADM_core/src/ADM_crashdump_mingw.cpp
+++ b/avidemux_core/ADM_core/src/ADM_crashdump_mingw.cpp
@@ -376,6 +376,6 @@ LONG WINAPI TopLevelExceptionHandler(struct _EXCEPTION_POINTERS *exceptionInfo)
 void ADM_backTrack(const char *info, int lineno, const char *file)
 {	
     char title[2048]={0};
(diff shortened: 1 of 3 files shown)


eumagga0x2a:
I hope it is tolerable to bundle both unrelated commits improving merely the legibility and the visuals of the console output in a single pull request. The second patch addresses the issue with `ADM_{info,warning,error}` adding a space character to the buffer. Present since [[Core] Add colored funcs ADM_warning, ADM_info and ADM_error](https://github.com/mean00/avidemux2/commit/3b3260653605c569909b133c9b085a3f93d864a6), it resulted in the space character being printed _past_ the closing newline at the line with the following message, destroying alignment.

The most confusing effect: run `avidemux3_qt5` from terminal and close the Avidemux window via Alt+F4. The last message (usually from `abortExitHandler`) in terminal is formatted with `ADM_info` resulting in the prompt being pushed one column to the right. Now scroll through the command history list: the space between the prompt and following command in the console has gone.

PR #108, opened 2017-03-10
[prefs] Add optional alternative keyboard shortcuts (incomplete)

eumagga0x2a:
This is an experimental and incomplete minimally invasive patch to add optional laptop-friendly keyboard shortcuts. It doesn't address hardcoded tooltips in the navigation widget yet. Toggling the preference requires a restart of Avidemux for the changes to take effect.


mean00:
If we go down that road, why no make these configurable ?
i.e. each is a triplet
Description / Actual Key / display shortcut

rojagit:
Will this take care of bringing back [ and ] ? Life is miserable without it

eumagga0x2a:
> If we go down that road, why no make these configurable ?
> i.e. each is a triplet
> Description / Actual Key / display shortcut

I fear I don't quite get the idea. Would you prefer making every single shortcut a pref? Or every alternative shortcut a pref? Prefs expect value pairs (like `Action:shortCut`, so this is OK). There is no GUI to set them yet, encouraging users to edit `config3` manually is probably the easiest way to get [into trouble](https://github.com/mean00/avidemux2/pull/51). Only a small subset of shortcuts raises a controversy, this is why I wanted to avoid to touch the `MenuEntry` struct and `myMenu{Edit,Go}` vectors.

My immediate goal was to produce a cheap stop-gap solution (i.e. something still doable with my limited skills) to address usability issues with laptop keyboards. Qt allows to assign multiple shortcuts to an action, which I investigated and got basically working, but backed off from that path because of poor discoverability of secondary shortcuts.

> Will this take care of bringing back [ and ] ?

If we end up with user configurable keyboard shortcuts, you'll be able to assign these keys (which are horribly inconvenient on non-US keyboards and completely absent in some Russian layouts). The current patch suggests `I` and `O` for that.

mean00:
Basically enhance your patch by making it configurable
The default keys stays the same, but you can set an alternate shortcut, which is as below by default
and that you can reset

That way, even if you have US/russian/German keyboard and using desktop/laptop keyboard
everybody can tune it as it pleases

The pref would be a int, -1 would mean default, else the key to use
It's just that there will always be someone with a special keyboard/language that will be unhappy if not done that way

That would end up as
ACT_PREF_ALT_MARKA = 'I'
ACT_PREF_ALT_MARKB = -1 (default)

etc etc

This would then be constructed dynamically, no problem
That's why it's done that way, instead of relying to  Qt menu management

static const kbdAccel _extraKbdAccels[] = {
 +            {ACT_MarkA,        "I"},
 +            {ACT_MarkB,        "O"},
 +            {ACT_ResetMarkers, "R"},
 +            {ACT_GotoMarkA,    "A"},
 +            {ACT_GotoMarkB,    "B"},
 +            {ACT_Begin,        "H"},
 +            {ACT_End,          "E"}
 +        };
 +

mean00:
(Only a few Alternate shortcut could be tuned)

eumagga0x2a:
Thank you for the clarification. I'll see how far I can get with severe time shortage looming.

eumagga0x2a:
I'm closing this and opening a PR for another tentative approach.

PR #109, opened 2017-03-11
[UI] Enable or disable menu items dynamically

eumagga0x2a:
This patch cleans up the undo queue a bit, getting rid of global variables and providing necessary helper functions used to determine whether a particular menu item should be enabled or disabled. Disabling e.g. the items in the "Auto" menu avoids Avidemux crashing when selecting any of the items without a video loaded.

Disabling the "Save as Image" submenu instead of the submenu actions is clumsy, but I haven't found a solution for that during the time available.


eumagga0x2a:
I'm sorry, I forgot about disabling menus while playing (evaluating `playing` in `MainWindow::setMenuItemsEnabledState`). I would add it in a followup PR a bit later, if this one turns out to be acceptable.

mean00:
Seems calling from command line is not taken into account
i.e.
avidemux3_qt5 foobar.mkv will not unlock the menus

PR #114, opened 2017-03-15
[Qt] Fix regressions concerning menu items enabling and disabling

eumagga0x2a:
This patch fixes hopefully all the regressions resulting from the recent beautification and simplification of the `MainWindow::setMenuItemsEnabledState` haystack along with some cosmetic whitespace changes and a couple of comments clarifying my own misleading wording ;-)


mean00:
The simplify is not really about factorizing
It' s more about saying simply what one wants to do, rather than how he's doing it

eumagga0x2a:
Thank you, the word "factorize" evaporated from my brain at the very moment I needed it.

PR #115, opened 2017-03-16
[prefs/Qt] Add alternate configurable keyboard shortcuts (initial backend part)

eumagga0x2a:
Another rough try, this time the keyboard shortcuts are made configurable (no configuration frontend yet). Additionally, allow changing the semantics of Up and Down keys for navigation.

`Ctrl+Pg{Up,Down}` will keep working even with "alternate" shortcuts active.

This patch doesn't address hardcoded tooltips in the navigation widget and doesn't implement rebuilding menus on closing the preferences dialog, required to make the effect of toggling the keyboard shortcuts prefs immediate.


mean00:
Looks good, thanks

(When i see large chunk of switch/case code  with only constants
i'm hearing the "simplifier" rattling in its drawer)

PR #116, opened 2017-03-17
[UI] Update shortcuts at runtime, bundle populating of action lists in a separate function

eumagga0x2a:
This patch implements updating the tunable action shortcuts (just these) at runtime, removing the earlier requirement to restart Avidemux for the changes to take effect.

It moves also a little bit of the menu items enabling and disabling code around, which might be helpful if we need to rebuild menus at runtime later, requiring the action lists to be recreated.


eumagga0x2a:
Now with tooltips in the navigation and selection widgets reflecting the currently active shortcuts.

I'm not quite convinced that the shortcuts in the tooltips should be uppercase, but it would be trivial to remove `toUpper()` in the future.

PR #117, opened 2017-03-17
[Qt] Enable and disable navigation buttons dynamically

eumagga0x2a:
This patch implements enabling and disabling buttons in the navigation and selection widgets similar to menu items depending on playback state and loaded video.


eumagga0x2a:
If you generally don't like the view of the main window with greyed out buttons, I won't insist. This feeling was the reason why I didn't try to disable top-level menus even if inside all items were greyed out.

PR #123, opened 2017-03-19
[UI/Qt] Hide non-functional volume widget with PulseAudioSimple

eumagga0x2a:
On Linux, we default to PulseAudioSimple output, which doesn't support volume control. Showing users a non-operational volume slider may be confusing, hide the volume widget if PulseAudioSimple is selected.

Furthermore, reinterpret `UI_updateActionShortcut` as the general way to perform actions on saving preferences, rename in to `UI_applySettings`.


mean00:
I would prefer it to be a property of the audio device
i.e. hasVolumeControl()

That way it is enabled/disabled for all, e.g. win32 & mac too

PR #137, opened 2017-03-29
[main] Fix version string generation

eumagga0x2a:
I would restrain from pushing even trivial changes on my own while in release mode. This too-late-patch fixes the version string generation for the Avidemux log header.


mean00:
I'll just wait a day or two to see if people are reporting nasty problems and switch back to "normal" mode
+ create a support branch
Then you can merge all you want :)

eumagga0x2a:
There is no urge :-) I've deleted the second commit in this PR, the patch there worked, but the approach was wrong. It looks like the current ADM code confuses unsetting a variable (`ADM_SUBVERSION`) and setting it to zero. I'll fix it tomorrow.

eumagga0x2a:
The new version of the second patch unifies the logic of the version string generation (developer build vs. release) and takes care about more places in the tree.

eumagga0x2a:
Oops, did you mean that I should have merged these patches into the "support" branch first?

mean00:
No, support is for emergency fix on 2.6.19

PR #142, opened 2017-04-03
[muxer] Populate codecpar, silence lavc deprecated warning

eumagga0x2a:
This patch is probably total rubbish, as I don't understand the matter well enough. Of course, it silences the warning [seen by Jan](http://avidemux.org/smif/index.php/topic,17642.0.html); encoding with FFmpeg-based and other encoders as well as saving in copy mode still seems to work fine. Nevertheless, could you have a look, please?


mean00:
Looks ok, i missed that change set

eumagga0x2a:
Merci beaucoup

PR #144, opened 2017-04-18
[Qt/mainfilter] Disable horizontal scrolling to work around rendering issues

eumagga0x2a:
This one could be controversial because tooltips become the only way to access long filter descriptions which get cut off.

Enabling word wrap for labels doesn't work well for me because list items must be 4 lines high to accomodate some translations inc. the German one while `QFontMetrics::elidedText` fails to take `<small>` tags into account.

The following screenshots show the rendering issue.
video-filter-manager-redraw-issue

video-filter-manager-available-selection-extends-into-active


eumagga0x2a:
I think I've found a satisfactory solution, closing this PR.

PR #148, opened 2017-09-08
Allow to specify which port to use for the connection of vapoursynth proxy

Hinterwaeldlers:
Working in the same way, as the modification done for avisynth proxy


mean00:
thanks

PR #151, opened 2018-01-08
Update avidemux_ko.ts

sheppaul:
Fixed some translations

### DIFF
--- a/avidemux/qt4/i18n/avidemux_ko.ts
+++ b/avidemux/qt4/i18n/avidemux_ko.ts
@@ -945,11 +945,11 @@ Drop timing informations ?</source>
     <message>
         <location line="+10"/>
         <source>© 2001 - 2017  Mean</source>
-        <translation>© 2001 - 2017  Mean</translation>
+        <translation>© 2001 - 2018  Mean</translation>
     </message>
     <message>
         <source>© 2001 - 2016  Mean</source>
-        <translation>© 2001 - 2016  Mean</translation>
+        <translation>© 2001 - 2018  Mean</translation>
     </message>
     <message>
         <location line="+19"/>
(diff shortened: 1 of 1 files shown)


avidemux/qt4/i18n/avidemux_ko.ts
```diff
@@ -1940,7 +1940,7 @@ Drop timing informations ?</source>
     <message>
         <location line="+16"/>
         <source>Pr&amp;eferences</source>
-        <translation>환경 설정(&amp;E)</translation>
+        <translation>환경 설정</translation>
```
eumagga0x2a: Why do you want to remove the kbd accelerator? Does it look confusing? It works fine for me at least on Linux with a usual pc105 keyboard type in 한글 mode. Does it fail for you?

All other changes are OK for me, thank you.

avidemux/qt4/i18n/avidemux_ko.ts
```diff
@@ -945,11 +945,11 @@ Drop timing informations ?</source>
     <message>
         <location line="+10"/>
         <source>© 2001 - 2017  Mean</source>
-        <translation>© 2001 - 2017  Mean</translation>
+        <translation>© 2001 - 2018  Mean</translation>
```
eumagga0x2a: Please do not correct strings which are wrong but don't need to be translated via translation. They need to be fixed in the application. The right thing to do would be to replace `<translation>© 2001 - 2017  Mean</translation>` with `<translation></translation>`.

sheppaul: If it should be fixed in the application, I have no idea the reason why the string is in translation text though I don't really care about it.

avidemux/qt4/i18n/avidemux_ko.ts
sheppaul: See the attached screenshot. It is the only entry having a kbd accelerator. It just looks weird to me. I'll add it again someday if required. :p

avidemux/qt4/i18n/avidemux_ko.ts
eumagga0x2a: The string is picked up automatically while parsing avidemux/qt4/ADM_userInterfaces/ADM_dialog/about.ui where we have a few strings which really need translation, so excluding the entire file is not an option.

Translation files will be regenerated shortly before release and changes to this string in the source will invalidate existing translations.

avidemux/qt4/i18n/avidemux_ko.ts
eumagga0x2a: Well, all other actions except of 편집 초기화 (Reset Edit) are directly accessible via kbd shortcuts (on macOS, unlike other operating systems, preferences have the default kbd shortcut "Command + ," and are located in another menu) and preferences are a menu entry one uses quite often, methinks. There are accels in the Video menu too.

Anyway, I'll merge this as-is and leave accels to your discretion. Thank you for your contribution.

PR #154, opened 2018-02-10
PL translation update

warchlak:
w:2860/3458


eumagga0x2a:
Thank you. FYI, translation files will be updated shortly before the next release, there have been some changes in the code affecting translatable strings, e. g. the "Play filtered" checkbox has been turned into an action and lives now in the toolbar and in the "Video" menu. Current translation files don't reflect these changes yet.

PR #156, opened 2018-03-03
avsfilter bugfix

oshrk:
avsfilter does not work after wine git commit:
f16cf541068831074a6891688d947f0ce32abdd9 (kernel32: GetTempPathW must zero the remaining buffer.)
GetTempPathW receives size in bytes from wine_tmp_path, but it expects size in wchars.
GetTempPathW (wine source dlls/kernel32/path.c) crashes in line:
memset(path + ret, 0, (count - ret) * sizeof(WCHAR));
because of wrong count value.

### DIFF
--- a/addons/avsfilter/winetmppath.cpp
+++ b/addons/avsfilter/winetmppath.cpp
@@ -23,7 +23,7 @@
 extern "C" void wine_tmp_path (AVS_PIPES *avsp, int num)
 {
   WCHAR path[MAX_PATH];
-  GetTempPathW(sizeof(path), path);
+  GetTempPathW(sizeof(path)/sizeof(path[0]), path);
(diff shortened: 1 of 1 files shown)


mean00:
thanks

oshrk:
Could you please rebuild avsload.exe and pipe_source.dll and update
avidemux_plugins/ADM_videoFilters6/avsfilter/avsload_pipe_binary.tar.bz2 
Thanks

mean00:
Sure
Might take a few days

mean00:
[pipe_try1.tar.gz](https://github.com/mean00/avidemux2/files/1789744/pipe_try1.tar.gz)
Does this work ?
(beware it has been compiled with vc++ 2015, 32 bits mode)

PR #159, opened 2018-04-17
Sturtup > Load > Window->Maximize [bug?]

warchlak:
Scenario:
AppStart > Load Video [WindowState=Normal]
Maximize Button Click: video stays in original size [!!]
Now Restore > Maximize again and it works from now.

Im not familiar with code, but changing this line can fix this?


eumagga0x2a:
Thank you for your report, I agree that this behavior is confusing. However, the proposed patch won't help if the video has been played before maximizing the window. I'm going to fix it in changeEvent().

PR #160, opened 2018-05-04
PL translation update

warchlak:
u11 (x264/5)


eumagga0x2a:
Thank you.

By the way, it looks like initializing `ignoreResizeEvent` to `false` is not required anymore, so changing it to `true` would be at least harmless. It was a remnant of an early attempt to prevent small videos from being automatically upscaled when loaded. The issue was fixed by the part of [[GUI] Fix small videos changing zoom to fit into window on stopping playback](https://github.com/mean00/avidemux2/commit/0a16c2908abf7c5a5607cc006bf0843cf5488efd#diff-47b20687090bf006e82bdf5bd721630b) which set `ignoreResizeEvent` to `false` at the end of `MainWindow::setMenuItemsEnabledState`.

PR #161, opened 2018-06-01
Update avidemux2.desktop

eszlari:
This updates the file and makes it pass `$ desktop-file-validate avidemux2.desktop` without warnings.

By the way: Avidemux is now available as Flatpak on Flathub:
https://flathub.org/apps/details/org.avidemux.Avidemux
Build manifest:
https://github.com/flathub/org.avidemux.Avidemux

### DIFF
--- a/avidemux2.desktop
+++ b/avidemux2.desktop
@@ -1,11 +1,10 @@
 [Desktop Entry]
-Encoding=UTF-8
-Name=avidemux2
+Name=Avidemux
 GenericName=Video Editor
 Comment=Multiplatform video editor
-Exec=avidemux2_gtk
+Exec=avidemux3_qt5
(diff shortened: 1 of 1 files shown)


eumagga0x2a:
It is generally expected that packagers provide their own launcher, but having a valid template upstream is never wrong. Thank you for your contribution.

eszlari:
> It is generally expected that packagers provide their own launcher

I don't quite get what you mean by that?! This file 100% reusable for every Linux package, there is nothing flatpak specific about these modifications.

eumagga0x2a:
The upstream .desktop file was outdated and unused since the GTK GUI had been abandoned. People who packaged Avidemux for Linux distributions wrote their own .desktop files for avidemux3_qt5.

PR #162, opened 2018-06-03
add icon with better quality

eszlari:


eumagga0x2a:
Since a few months, Avidemux uses updated high-resolution application icons on Windows and macOS, albeit not on Linux (the problem was creating satisfactory low-resolution versions). I am not sure the Linux version should choose a totally different color palette and design language – so we need either to create a set of icons in usual sizes (16, 22, 24, 32, 48, 256) matching to some extent the appearance on other platforms or redo the application icons on Windows and macOS.
avidemux_icon_windows_256x256
avidemux_icon_osx_256x256

eszlari:
Are they in the repository? i can't find them.

eumagga0x2a:
The macOS icons file was last touched in [[macOS] Fix wrong inclination of stripes on the clapperboard in the application icon](https://github.com/mean00/avidemux2/commit/e9fc31c0a7d57fdb7dde97eb4e12c8d178f6c65c).

The high-resolution icon for Windows was added by [[Windows] Add high resolution, PNG compressed bitmap to the application icon](https://github.com/mean00/avidemux2/commit/b6242460fd8e41c9dc6150188c80a3acbbbf0ee4). The original SVG files which were used to produce the graphics are not in the repository.

eszlari:
Linux supports SVG icons, so please could you add them? Or at least a PNG? (I would prefer the macOS version for Linux)

eszlari:
ping?

eumagga0x2a:
A set of new Avidemux icons for Linux is definitely on radar but not a priority ATM. The hard part is designing low resolution and symbolic (monochrome) versions. Once this is accomplished and lands in a release, adding the source SVG might be reconsidered.

Personally, I am happy with `Icon=org.gnome.Totem` and `StartupWMClass=avidemux3_qt5` in the Avidemux .desktop file as a private temporary solution.

What concerns the proposed icon, thank you very much, but we won't take it upstream.

eszlari:
> What concerns the proposed icon, thank you very much, but we won't take it upstream.

Of course, I'm not asking you to do that, now that you shown that there already better ones available. Why can't you just replace the one at the top level with the macos icon?

eumagga0x2a:
> Why can't you just replace the one at the top level with the macos icon?

Because the macOS icon is just for macOS, it doesn't fit into other environments and visual styles.

eszlari:
> Because the macOS icon is just for macOS, it doesn't fit into other environments and visual styles.

This might be the case, but the icon is definitely much better than the one that is at the top level at the moment.

PR #166, opened 2018-06-09
Video "Copy" text translation

warchlak:
Issue: Copy text in output video combobox (and audio) is replaced by english name.

I'm not sure if this is a good place to implement this.
Because display name is used by functions, the translation should be global (right?).

### DIFF
--- a/avidemux/common/ADM_videoEncoder/src/ADM_dynVideoEncoder.cpp
+++ b/avidemux/common/ADM_videoEncoder/src/ADM_dynVideoEncoder.cpp
@@ -26,8 +26,8 @@
 static int currentVideoCodec=0;
 
 ADM_videoEncoderDesc copyDesc={
-        "Copy",
-        "Copy",
+        QT_TRANSLATE_NOOP("MainWindow", "Copy"), // name
+        QT_TRANSLATE_NOOP("MainWindow", "Copy"), // tag - unused i guess
         "Copy encoder",
         ADM_VIDEO_ENCODER_API_VERSION, //uint32_t     apiVersion;            // const
 


eumagga0x2a:
The structure ADM_videoEncoderDesc is defined in avidemux_core/ADM_coreVideoEncoder/include/ADM_coreVideoEncoderInternal.h, so the first member is the codec tag, which must be unique and is how a codec must be called in scripts. This means, we can't make the tag translatable, otherwise scripts would need to be rewritten for each locale which provides a translation.

The second member is what gets shown in the menu. Frankly speaking, even if it works, I don't like the idea to make codec names translatable (don't forget all the other codecs in avidemux_plugins/ADM_videoEncoder). Apart from making user support harder, it would result in the menu being sorted differently in each locale.

The real issue is that we have that default text in gui2.ui which lands in translation files and seduces translators to translate what doesn't need to be translated at all, while we can't add a comment to warn them.

warchlak:
I think you misunderstood me.
Im not trying to translate codec names, only that one named Copy. Proposed changes are not doing it?

If they do translation function should use gui translation for that text (i guess, have no idea how it works) so translation is not unnecessary.

eumagga0x2a:
"Copy" **is** a codec (pseudo-codec) name. IMHO, it is the same sort of a technical term like "Mpeg4 AVC (x264)". And no, the patch doesn't work. You could replace "Copy" with a translation in `setupMenus()` instead:

```
diff --git a/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp b/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
index 129b6b0f..cb383806 100644
--- a/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
@@ -1893,6 +1893,8 @@ void setupMenus(void)
     for(uint32_t i=0;i<nbVid;i++)
     {
         ADM_ve6_getEncoderInfo(i,&name,&maj,&mn,&pa);
+        if(!strcmp(name,"Copy"))
+            name=QT_TRANSLATE_NOOP("MainWindow","Copy");
         WIDGET(comboBoxVideo)->addItem(name);
     }
 
@@ -1906,6 +1908,8 @@ void setupMenus(void)
     for(uint32_t i=0;i<nbAud;i++)
     {
         name=audioEncoderGetDisplayName(i);
+        if(!strcmp(name,"Copy"))
+            name=QT_TRANSLATE_NOOP("MainWindow","Copy");
         WIDGET(comboBoxAudio)->addItem(name);
     }
```
(both for video and audio). I am not convinced that we need that.

warchlak:
You are a tough negotiator.
Okay, i'm leaving it alone... for now ;)

PR #173, opened 2019-01-04
rename Linux metadata

eszlari:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html


eumagga0x2a:
Thank you. Would you mind taking the route with the [<launchable/> ](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html#tag-dapp-launchable) tag thus avoiding renaming the .desktop file, please? Like vlc does in their vlc.appdata.xml.

Renaming the launcher means annoying at least the ArchLinux package maintainer for zero benefit.

eszlari:
I was planning on doing a follow-up PR, to have these files (and a icon) installed by cmake. So everything should work out of the box.

eumagga0x2a:
Thank you, I know they want us do these strange things.

There is no suitable icon set to install yet, unfortunately. I hope to manage to do one, albeit not in time for the next release at least.

PR #175, opened 2019-01-19
linux/cmake: install xdg metadata

eszlari:


eumagga0x2a:
Thank you.

eszlari:
Would you update the release entry in appdata.xml, whenever you make new releases?

It would be also nice, if you could make a new release soon. I would like to update the flatpak (https://flathub.org/apps/details/org.avidemux.Avidemux) with Qt 5.12 support and these metadata fixes.

PR #176, opened 2019-01-19
translation pl update

warchlak:


eumagga0x2a:
Thank you very much. We have a small problem that the patch has been created based on git master. However, the current work is done on the ffmpeg4x branch which contains updated translation files. Would it be doable for you to resubmit this translation update based on the ffmpeg4x branch instead of master?

If this is too complicated, I could try to pull it and see whether I manage to resolve resulting merge conflicts when merging master into ffmpeg4x.

eumagga0x2a:
> You have cleared the file from old translations, so the new one is much smaller

Not just that. Recent modificatons of the subtitle plugin required manual interventions to preserve existing translations despite source string changes.

> I'll use the dictionary from the old file and prepare a new one.

This would be very nice, thank you.

eumagga0x2a:
I noticed that you modified the qtbase translation as well. These translations are grabbed from qt upstream once Qt in our build nodes is updated. It would be far better to submit the changes upstream as downstream changes would get lost in the course of the next update.

If the work on the ffmpeg4x version of the patch for avidemux_pl.ts hasn't progressed much yet, I would merge this into master and deal with merge conflicts when merging master into ffmpeg4x like in the case of the Greek translation update as the beta release is imminent.

eumagga0x2a:
Merging, thank you for your contribution.

warchlak:
This is too complicated for me. I'll deal with qtbase problem later.

Second, more important thing. Are you suggesting that I should not update the translation and wait until you have merged the branches? I got lost a bit and I do not know what to do.

eumagga0x2a:
Everything is already accomplished, thank you. The pull request has been merged into master and master pro forma merged* into ffmpeg4x which is the current beta release branch for 64 bits builds for Linux, macOS and Windows 7+.

The updated avidemux_pl.ts from ffmpeg4x has been subsequently copied over to the legacy-compat branch which is the source of 32 bits builds trying to keep compatibility with Windows XP (which refuses to die).

*) I simply applied the patch to ffmpeg4x branch and let git figure out the diff necessary to resolve the conflict.

Regarding qtbase translations, it would be better to submit patches where these files come from – https://code.qt.io/cgit/qt/qttranslations.git/ –  following https://wiki.qt.io/Qt_Localization because any modifications to these files in the Avidemux codebase will be overwritten when the translations get refreshed (copied over from upstream).

I apologize for an accidental mixup of English and German while writing "Polish" in two of my last commits on the ffmpeg4x and legacy-compat branches.

PR #178, opened 2019-03-20
avidemux 2.72

twinl40:
good afternoon how can you do that when you convert a mkv file stay with audio output aac fdk 
how I set it


eumagga0x2a:
Please use the [forum](https://avidemux.org/smif/index.php) for support requests.

PR #179, opened 2019-08-14
[cmake] Fix library permissions

steils:
See also:
https://avidemux.org/smif/index.php?topic=16522.0
https://bugs.archlinux.org/task/56373


mean00:
Thanks

PR #180, opened 2019-08-19
Blend plugin

krakow10:
Plugin in videoFilters6/blend to average chunks of sequential frames to create a high quality timelapse effect.  I have been using it to create motion blur by speeding up slow motion footage of video games which have no motion blur.  Also hevc_nvenc maximum bitrate is too low :)


mean00:
Thanks
It looks like could be improved in terms of speed 
(i've taken also the change to ffNnvencHEVC, please split patches on different topics)

PR #181, opened 2019-11-12
cmake: fix detection of CPUs

jirislaby:
The code defines a function foo returning int, but contains no return.
So the detection fails with "CPU not supported". As new gccs warng abotu
missing return, it causes the detaction to fail. Switch foo to return
void.


mean00:
Thanks

PR #182, opened 2019-11-12
[ivtcDupeRemover] return true from ivtcDupeRemover::getNextFrame

jirislaby:
It is declared as returning a bool, but returns nothing. But the callers
expect something to be returned: true in case it succeeded.

**Double check I got it right...**

### DIFF
--- a/avidemux_core/ADM_core/include/ADM_core6_export.h
+++ b/avidemux_core/ADM_core/include/ADM_core6_export.h
@@ -50,4 +50,10 @@
 # define ADM_CORE6_NO_DEPRECATED
 #endif
 
+#ifdef __GNUC__
+#  define ADM_CORE6_NORETURN __attribute__((noreturn))
+#elif defined(_WIN32)
+#  define ADM_CORE6_NORETURN __declspec(noreturn)
+#endif
+
 #endif
--- a/avidemux_core/ADM_core/include/ADM_crashdump.h
+++ b/avidemux_core/ADM_core/include/ADM_crashdump.h
@@ -17,7 +17,7 @@ extern "C"
 	typedef void ADM_fatalFunction(const char *title, const char *info);
         typedef void ADM_sigIntFunction(void);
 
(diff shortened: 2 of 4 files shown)


eumagga0x2a:
Could you please explain under which circumstances the added return statement can be ever reached?

jirislaby:
@eumagga0x2a compiler warns about the `default` label. `ADM_assert` expands to `ADM_backTrack`, but the latter is not defined with `no_return` attribute. And it actually can return on some platforms (HAIKU or sun), see: https://github.com/mean00/avidemux2/blob/59c11947d07fead588336a4b75a73fc2681711a0/avidemux_core/ADM_core/src/ADM_crashdump_unix.cpp#L185-L231

jirislaby:
Based on the previous comment, it should `return false`, given there is `return true` hidden in the `END_PROCESS` macros used there.

mm1044:
> the latter is not defined with no_return attribute

Maybe that is the root issue?
Why are these two cases "unsupported"? Unwanted, impossible or ToDo?
Shouldn't `ADM_backTrack` print an error and exit, instead of ignoring and returning?

> it actually can return on some platforms (HAIKU or sun)

You may want to add a comment here about that.

jirislaby:
> > the latter is not defined with no_return attribute
> 
> Maybe that is the root issue?
> Why are these two cases "unsupported"? Unwanted, impossible or ToDo?
> Shouldn't `ADM_backTrack` print an error and exit, instead of ignoring and returning?

IMO, it should always die for consistency (and be declared as `noreturn`). But this is the first time I looked into the avidemux code so I am not the one to decide this...

> > it actually can return on some platforms (HAIKU or sun)
> 
> You may want to add a comment here about that.

What do you mean by _here_ here?

Anyway, pushing implementation of the above shortly.

mm1044:
> What do you mean by `here` here?

I meant at `ivtcDupeRemover::getNextFrame()`,
but that is superseded by your new/current commit.

PR #187, opened 2020-07-24
add support for alternate key for Delete action

sandsmark:


eumagga0x2a:
Thank you, the changes as such look great, and yes, all the magic numbers in menu action enabling / disabling code is an unmanageable hell in need for an urgent fix.

When merged, the patch will finally solve the problem on OSes other than macOS with keyboards without a Delete key. In return, it will require users to edit the configuration file (config3) to keep their physical Delete key (or the Backspace key on macOS) working as expected while enjoying the comfort of alternative keyboard shortcuts.

Would you please agree to set the default alternative shortcut for delete instead of to "K" to the current normal default value? This will allow all users who need to work around a hardware limitation or want to use a custom shortcut like "K" to accomplish it while letting the majority (IMVHO) unaffected.

eumagga0x2a:
I'd like to clarify that I don't mean that more code changes need to be added to this PR, I would merge it as-is, changing the mentioned detail later.

eumagga0x2a:
Thank you for your contribution.

sandsmark:
sorry, I didn't have time to follow up, but thanks.

I'll see if I find some time to start replacing the numbers with enums representing the same numbers (as a first step at least), if you think that's a viable way forward.

eumagga0x2a:
Thank you, I had something similar to `MainWindow::searchMenu` in mind to return position of QAction in given menu matching given Action member of MenuEntry struct (avidemux/common/myOwnMenu.h:27) from the avidemux/common/gui_action.names enum.

I fear that introducing one more independent enum / define layer would merely shift the maintainability burden.

PR #188, opened 2020-10-06
X265 options

xyzzy42:
Here are a couple options for x265 that seem useful.

I'd like to be able to duplicate at least the slow preset from x265 in the advanced config, so that it's possible to use slow + extra options.  If I want a longer keyframe interval for 60fps content it's necessary to give up some of options that the x265 authors think are useful enough to include in the preset.


eumagga0x2a:
Thank you very much.

PR #189, opened 2020-10-07
X265 options

xyzzy42:
More options for x265 to allow at least the slow preset to be configured.  There are still some more to fully get slow, and tons more total, but I think probably the options used in the presets area really worth adding.

I've tested this with x265 3.3 and they appear to work as expected.  I think all these options have been x265 for a number a versions, but I haven't gone back and found out how long.

### DIFF
--- a/avidemux_plugins/ADM_videoEncoder/x265/ADM_x265.h
+++ b/avidemux_plugins/ADM_videoEncoder/x265/ADM_x265.h
@@ -73,6 +73,7 @@ extern "C"
     3, /*    uint32_t me_method */ \
     16, /*   uint32_t me_range */ \
     5, /*    uint32_t subpel_refine */ \
+    0x03, /* uint32_t limit_refs */ \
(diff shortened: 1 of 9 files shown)


eumagga0x2a:
Great, thanks. All options added here were introduced in x265 at or before v1.6 and therefore safe, we need only to be careful about stuff introduced after v2.4.

xyzzy42:
Release 2.4 of x265 was `X265_BUILD` API of 116.  I see a bunch of stuff in the avidemux code for APIs in the 30s and 40s, around release 1.4 and 1.5.

Would it be safe to delete that?  Some of it is adds UI elements that are then disabled, so the config dialog could be cleaned up to remove obsolete settings.

eumagga0x2a:
Yes, please, I assume it is safe to remove that. Anything older than the libx265 version in the oldest supported Debian release (currently 95) doesn't need to be taken into account, IMVHO. Official Avidemux appImages bundle the version 146.

PR #190, opened 2020-10-08
X265 fix badapt

xyzzy42:
Fixed b-adapt not setting correctly.  The trellis quant stuff was leftover from x264.  I had wondered what that option did...


eumagga0x2a:
Thank you very much for the fix.

PR #191, opened 2020-10-08
X265 options

xyzzy42:
rdoq and psy-rdoq.  Almost done.  Just limit-modes is left.

I tried to copy the tab/space style of the surrounding code, but the existing code is not very consistent, so maybe some of the code isn't using the right one.


eumagga0x2a:
> Almost done. Just limit-modes is left.

Great, thanks, I'll wait with syncing translation files until you give me a sign then.

> I tried to copy the tab/space style of the surrounding code, but the existing code is not very consistent, so maybe some of the code isn't using the right one.

I fear there is simply no style consistency in vast areas of the historically grown codebase whatsoever. On my part, I just try to keep content and extensive whitespace / style changes separate, if possible.

For new code which doesn't mean to blend into close environment, tabs are not used, scope levels are idented by 4 spaces and the opening curly bracket is placed on a new line. I am not sure there is much else left in the sense of visual style.

PR #192, opened 2020-10-08
X265 options

xyzzy42:
I think this is everything to allow one to do the "slow" preset.  I removed some old 1.5 options.


eumagga0x2a:
Thank you.

PR #193, opened 2020-10-13
[x265] Store data with bit depth combo box items

xyzzy42:
This avoids using a bunch of loops to find the combo box item as a string in listOfBitDepths to look up the depth.  It's just one line now.   Also handles the case where a saved project uses an unsupported depth.

### DIFF
--- a/avidemux_plugins/ADM_videoEncoder/x265/qt4/Q_x265.cpp
+++ b/avidemux_plugins/ADM_videoEncoder/x265/qt4/Q_x265.cpp
@@ -222,12 +222,12 @@ x265Dialog::x265Dialog(QWidget *parent, void *param) : QDialog(parent)
 
         QComboBox *depths=ui.comboBoxBitDepth;
         depths->clear();
-        depths->addItem(QString(automatic));
+        depths->addItem(QString(automatic), QVariant(0U));
         for(int i=0;i<NB_BITS;i++)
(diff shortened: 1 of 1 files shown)


eumagga0x2a:
Fantastic, just not ready to break Qt4 compatibility right now (don't think that Qt 5 < 5.2 matters). Could you please keep the old implementation for `QT_VERSION < QT_VERSION_CHECK(5,2,0)` ?

xyzzy42:
What version of Qt should be supported?  I think most of this doesn't need Qt5.2.

eumagga0x2a:
Please disregard my last comment, I somehow missed that `QVariant::QVariant(uint val)` is in Qt 4.8, so everything should be fine.

xyzzy42:
Hmm, I thought `QComboBox::findData()` was new, but it's 4.8 too.

eumagga0x2a:
The build with Qt4 does fail at

`myCopy.general.output_bit_depth = ui.comboBoxBitDepth->currentData().toUInt();`

as QComboBox in Qt 4.8 doesn't contain this method.

eumagga0x2a:
`myCopy.general.output_bit_depth = ui.comboBoxBitDepth->itemData(ui.comboBoxBitDepth->currentIndex()).toUInt();`

might work.

PR #194, opened 2020-10-14
Use data for the combo boxes

xyzzy42:
Storing a data value with each combo box item makes things a lot easier.

### DIFF
--- a/avidemux_plugins/ADM_videoEncoder/x265/qt4/Q_x265.cpp
+++ b/avidemux_plugins/ADM_videoEncoder/x265/qt4/Q_x265.cpp
@@ -79,19 +79,13 @@ static const aspectRatio predefinedARs[]={
 
 #define NB_SAR sizeof(predefinedARs)/sizeof(aspectRatio)
 
-typedef struct
-{
-    int bdValue;
-    const char *bdString;
-}bdToken;
-
-static const bdToken listOfBitDepths[]={
+static const idcToken listOfBitDepths[]={
     {8,"8"},
     {10,"10"},
     {12,"12"}
 };
 
-#define NB_BITS sizeof(listOfBitDepths)/sizeof(bdToken)
+#define NB_BITS sizeof(listOfBitDepths)/sizeof(*listOfBitDepths)
 
 static const char* listOfPresets[] = { "ultrafast", "superfast", "veryfast", "faster", "fast", "medium", "slow", "slower", "veryslow", "placebo" };
 #define NB_PRESET sizeof(listOfPresets)/sizeof(char*)
@@ -225,9 +219,9 @@ x265Dialog::x265Dialog(QWidget *parent, void *param) : QDialog(parent)
         depths->addItem(QString(automatic), QVariant(0U));
         for(int i=0;i<NB_BITS;i++)
         {
-            const bdToken *t = listOfBitDepths + i;
-            if(x265ProbeBitDepth(t->bdValue))
-                depths->addItem(QString(t->bdString), QVariant(t->bdValue));
+            const idcToken *t = listOfBitDepths + i;
+            if(x265ProbeBitDepth(t->idcValue))
+                depths->addItem(QString(t->idcString), QVariant(t->idcValue));
         }
 
         upload();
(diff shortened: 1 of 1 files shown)


avidemux_plugins/ADM_videoEncoder/x265/qt4/Q_x265.cpp
```diff
@@ -79,19 +79,13 @@ static const aspectRatio predefinedARs[]={
...
-
-static const bdToken listOfBitDepths[]={
+static const idcToken listOfBitDepths[]={
     {8,"8"},
     {10,"10"},
     {12,"12"}
 };
 
-#define NB_BITS sizeof(listOfBitDepths)/sizeof(bdToken)
+#define NB_BITS sizeof(listOfBitDepths)/sizeof(*listOfBitDepths)
```
xyzzy42, line 88: Using *listOfBitDepths will produce the right value if listOfBitDepths changes type.  There's a hard to find bug where someone forgets to change NB_XXX after changing the array type, but the code still compiles without any error or warning, and maybe works fine, until later when someone changes one of the types and the size becomes wrong.

avidemux_plugins/ADM_videoEncoder/x265/qt4/Q_x265.cpp
```diff
@@ -122,6 +122,60 @@ bool x265_ui(x265_settings *settings)
...
+    \param [in] initial An option first item to add.
+    \param [in] initdata The data to be assigned with optional first item, default is empty.
+*/
+template <typename ContainerT>
+static void fillComboBoxData(QComboBox *box, const ContainerT& items, const char* initial = NULL, const QVariant& initdata = QVariant())
+{
+    box->clear();
+    if (initial)
+        box->addItem(QString::fromUtf8(initial), initdata);
+    for (const idcToken *t = std::begin(items); t != std::end(items); t++)
```
xyzzy42, line 154: In C++11, I would write `const auto iter = std::begin(items)` and then it wouldn't matter kind of container or exactly what type was in the container.   One could call it with an array of {std::string, int} and again with a std::vector of {QString, unsigned long}, and it would work with both.  But making that work in C++98 is just too much of a pain.

avidemux_plugins/ADM_videoEncoder/x265/qt4/Q_x265.cpp
```diff
@@ -79,28 +79,22 @@ static const aspectRatio predefinedARs[]={
...
 
 static const char* listOfPresets[] = { "ultrafast", "superfast", "veryfast", "faster", "fast", "medium", "slow", "slower", "veryslow", "placebo" };
 #define NB_PRESET sizeof(listOfPresets)/sizeof(char*)
 
+static const char* listOfTunings[] = { "psnr", "ssim", "grain", "zerolatency", "fastdecode",
 #if X265_BUILD > 173
-static const char* listOfTunings[] = { "none", "psnr", "ssim", "grain", "zerolatency", "fastdecode", "animation" };
-#else
-static const char* listOfTunings[] = { "none", "psnr", "ssim", "grain", "zerolatency", "fastdecode" };
+    "animation"
```
eumagga0x2a, line 95: Are you sure this will work as expected with VC++? I had to do 1a3ab8a1f27f94ddbbc9f7556863f9bff8f56bdd in a similar situation.

Official Avidemux builds for Windows are built with Visual Studio, I haven't mastered the laborious setup of the build environment yet.

But this can be trivially changed later, after the PR is merged as-is.

avidemux_plugins/ADM_videoEncoder/x265/qt4/Q_x265.cpp
eumagga0x2a: As far as I can tell, there is no problem with using C++11 features, it is just necessary to explicitly set `-std=c++11` else Apple clang doesn't play nice.

avidemux_plugins/ADM_videoEncoder/x265/qt4/Q_x265.cpp
xyzzy42: It should work.  It's a pretty common thing and totally within the standard.

I think the issue you see in 1a3ab8a is that it the ifdef is inside the invocation of a preprocessor macro.  That's not allowed.  The preprocessor is a simple one pass processor, so it isn't meant to processor the ifdefs inside a macro invocation, then expand the macro.  See https://stackoverflow.com/questions/39534609/is-it-wrong-to-add-preprocessor-directives-in-a-function-like-macro

This is just a variable initialization in plain C, so there is nothing here that will break VC++.

eumagga0x2a: > I think the issue you see in 1a3ab8a is that it the ifdef is inside the invocation of a preprocessor macro.

AFAIR, I had to abandon preprocessor macros in argument lists of functions like `void GUI_Error_HIG(const char *primary, const char *secondary_format, ...);` and in definition of `diaMenuEntry streamOpt[]` due to broken functionality, but maybe I misattributed the problem.

Thank you very much for the patch and excellent documentation.

PR #195, opened 2020-10-25
X265 color

xyzzy42:
Add the ability to select the color primaries, color matrix, and transfer characteristics.  This was already in the Output 2 tab, but was disabled and never completed.

This will only be added to the bitstream if any of options are changed from "Unknown".

I've tried to sort the available options in a useful and consistent way and also indicate what kind of material will use them.  "Unknown" is translatable, but the other items aren't.  I omitted the reserved codes.

The UI is reorganized significantly for this.  Create new tab "Display" and put the PAR and color options here.  Then combine what was left of the Output 1 and Output 2 tabs into one Output tab.  That tab remains disabled.  The Video Format combo box also remains disabled on the Display tab.

Fix the PAR options so that selecting unspecified will leave the PAR unspecified, predefined will select one of the H.265 predefined SAR codes, and custom will select an extended SAR.

The predefined aspect ratios are now the ones defined in H.265.  The PAR values previously in the combo box are not in the standard and they are (arguably) incorrect.

I've tested with ffprobe -show_stream and it finds the indicated color information.  The PAR values work too, but will come from the container first, and only if no container is used, from the video stream.


eumagga0x2a:
Great, thank you very much for this contribution.

> The predefined aspect ratios are now the ones defined in H.265. The PAR values previously in the combo box are not in the standard and they are (arguably) incorrect.

The PAR values for PAL in x265_sar_names are unfortunately wrong for the vast majority of SD broadcasts and video DVD sources at least in Germany where the 720x576 resolution is predominantly used. No deal breaker anyway.

How does the added support for color primaries, matrices and transfer characteristics in the x265 plugin correspond to the current state of (in)ability to handle these (especially sorely lacking in case of BT.2020) in Avidemux? Does it depend on internal color handling in Avidemux getting fixed / redesigned to become really useful?

xyzzy42:
The PARs of 12:11 or 16:11 are correct for 720x576.  The width of 720 includes overscan.  The displayed width, the part of the screen which should make up the 4:3 or 16:9 image, is only ~704 of those pixels.
This can been seen in the H263, H264, and H265 specs, where the PAR 12:11 example is 720x576 with overscan, and the 24:12 example is the 352x576 without overscan.
The analog outputs of DVD players would almost always get this correct, with 704 and 720 width DVDs having the same PAR and 720 width DVDs having effectively 16 pixel wider image.
Digital display of DVDs was quite often wrong, treating the 720 width image as having no overscan and producing the incorrect 16:15 and 64:45 PARs.
If the source was originally analog video, then 16:15 is almost certainly wrong and 12:11 is correct.
But if someone scaled down a bluray rip from 1080p to DVD res, then they probably took the entire 1920x1080 DAR 16:9 image and converted into into a 720x576 DAR 16:9 image with no overscan, and a PAR of 64:25 is correct.  They should have used 704x576, but of course 720 is more than 704 so no one does this.

xyzzy42:
My understand of the colorspace settings, is that as long as the re-encoding doesn't convert to YUV->RGB->YUV, and it shouldn't, then it will be ok if the correct output is chosen.  If I know my source if BT.709 and select BT.709 as the output, it should be correct.  If I select the wrong output colorspace, then it will be incorrect.

If a filter does a conversion to/from RGB, then it should use the color matrix values to do that correctly.  I'm doubt any of them do.  However, the display driver might support this!

My understanding is the color primaries and EOTF are ignored by anything one might use to display the video.  It's just information a profession film editor might use.  The color matrix value is used by better display devices and software.  Or it's ignored and width of the video is used to select SMPTE 170M or BT.709.

PR #196, opened 2020-11-15
Allow choice of executable in AppImage

db-inf:
1. Because of the patch with sed in deployBusterMinimal.sh, to replace all absolute paths to `/usr` by relative paths `././` and thelike, AppRunBuster has to change directory to `"${HERE}/usr"`, which it indeed does. But then `"$@"` should be run through a loop to replace relative path or filename arguments in it by their `"realpath"`.

2. The new AppImage runtime passes a variable `ARGV0` to AppRun, which is set to the name/path used to execute the AppImage (for example the name of a symbolic link to the actual AppImage file). That way, if a user or his admin makes three symbolic links to the appImage file, let's say as `avidemux`, `avidemux_jobs`, and `avidemux_cli` (with or without versions in the name), the AppRun script can detect the suffix used, and call the corresponding binary.

3. This pull request should go with an instruction to the AppImage user, or his admin, to place symbolic links to the AppImage file. If the AppImage file is executed without symbolic link, avidemux3_qt5 is executed. If the symbolic link contains the substring "_jobs", avidemux3_jobs_qt5 is executed; else if the symbolic link contains the substring "_cli", avidemux3_cli is executed; else avidemux3_qt5 is executed.

### DIFF
--- a/appImage/AppRunBuster
+++ b/appImage/AppRunBuster
@@ -16,6 +16,14 @@ then
     exit 1
 fi
 
+# make relative-path arguments absolute before changing directory
+args=("--portable");
+paramprefix="--";
+for arg in "$@";
+do # add argument as is, or realpath of argument, depending on paramprefix
+	[ "${arg::2}" != "$paramprefix" -o -e "$arg" ] && args+=("$(realpath -- "$arg")") || args+=("$arg");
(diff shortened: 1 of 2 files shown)


mean00:
Thanks a lot

PR #197, opened 2020-11-26
[demuxers/MpegTS] Fix for 60000/1001 framerate

xyzzy42:
The ts.idx2 file stores framerate as fps * 1000, instead of a rational
number with an integer number and denominator, which would be much
better.

It must then convert that back into a rational so it can put inside a
fake AVI header.

It does not use a closest rational approximation algorithm but just hard
codes a few different common frame rates.  And the standard ATSC rate of
60000/1001 was not in the list.  So add it.

### DIFF
--- a/avidemux_plugins/ADM_demuxers/MpegTS/ADM_tsReadIndex.cpp
+++ b/avidemux_plugins/ADM_demuxers/MpegTS/ADM_tsReadIndex.cpp
@@ -285,6 +285,10 @@ bool    tsHeader::readVideo(indexFile *index)
             _videostream.dwScale=1001;
             _videostream.dwRate=30000;
             break;
+	case 59940:
+            _videostream.dwScale=1001;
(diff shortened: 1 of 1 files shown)


eumagga0x2a:
Thank you, missed that one.

Using a rational and keeping timebase and fps separate would be better, of course. I don't think that the current = legacy approach would help with timebase and fps changing midstream much anyway.

eumagga0x2a:
I'm sorry for misusing this old closed pull request as communication channel for an OT question, but as Avidemux 2.7.7 is about to be released, would you like to be mentioned in the release changelog?

And if yes, by name or by nick?

If the latter, linking to your GitHub profile would be undesirable, I guess?

xyzzy42:
Hadn't noticed there's no DM on github.  Thanks for putting me in the changelog.  I'd be happy to be mentioned by name and/or github profile link.

PR #198, opened 2021-01-07
Increased editor cache size default and upper limit

szlldm:
Hi!

Editing high demanding contents (eg. 4k 60fps or more) is very painfull, especially when searching for cutting point in the video.
Due to the default (and until now upper limit) 16 frames editor cache size, while navigating in the video, it will constantly re decode the video frames, therefore it cause a major slowdown during editing.

In this commit i increased the upper limit to 255 frames, and the default value to 60 frames.

### DIFF
--- a/avidemux/common/ADM_commonUI/DIA_prefs.cpp
+++ b/avidemux/common/ADM_commonUI/DIA_prefs.cpp
@@ -315,7 +315,7 @@ std::string currentSdlDriver=getSdlDriverName();
         multiLoadUseCustomFragmentSize.link(1,&multiLoadCustomFragmentSize);
 
         diaElemFrame frameCache(QT_TRANSLATE_NOOP("adm","Caching of decoded pictures"));
-        diaElemUInteger cacheSize(&editor_cache_size,QT_TRANSLATE_NOOP("adm","_Cache size:"),8,16);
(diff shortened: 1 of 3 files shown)


eumagga0x2a:
Thank you, let us do some maths.

The minimum amount of memory necessary to store a 3840x2160 YUV picture with 4:2:0 chroma subsampling is 3840 * 2160 * 3 / 2 = 12441600 bytes.

With hardware acceleration enabled, 12441600 * 60 ≅ 712 MiB which the graphics card needs to be able to allocate in its memory available for video decoding acceleration (maybe it can use the entire graphics memory and swap into the main one, I don't know, this may depend on hardware, operating system, driver and driver settings).

With editor cache size at maximum, a graphics card needs to be able to provide almost 3 GiB of video decode acceleration memory.

Fianlly, we should not forget that this is _per loaded video_, i.e. appending a second video doubles the requirements.

How should this scale? My experience with hitting the ceiling of available video surfaces was not good so far. Actually, the current number of 16 pictures in cache is already pretty dangerous in absence of a safe and reliable way to detect the actual limit for the particular system (we should not forget that other applications may want to allocate some video surfaces as well).

szlldm:
OK, i understand the reasoning. Would it be acceptable, if i keep the increased upper limit while reset the default value back to 16?

szlldm:
I think i will keep this changes in local.

PR #199, opened 2021-01-11
Crop filter keep aspect ratio option

szlldm:
Hi!

Added a "Keep aspect ratio" option to the Crop video filter.
When enabled, the aspect ratio of the source can be maintained on the selected crop area, if the rubber-band is dragged by the bottom-right corner. The top-left corner continue to be free dragging,the numeric inputs are unchanged.

(Use-case: recently i had to cut several scenes out of a video, and the areas of interest had different sizes and positions. )

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/crop/ADM_vidCrop.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/crop/ADM_vidCrop.cpp
@@ -74,6 +74,7 @@ CropFilter::CropFilter(ADM_coreVideoFilter *in,CONFcouple *couples) :ADM_coreVid
             configuration.left=0;
             configuration.right=0;
             configuration.rubber_is_hidden=false;
+            configuration.keep_aspect=false;
         }
         if(  in->getInfo()->width<(configuration.right+configuration.left))
                 {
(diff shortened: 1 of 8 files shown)


eumagga0x2a:
Thank you, the assumption that the source already has the desired aspect ratio may not hold in real life, but that could be addressed later, maybe by adding a menu which would include also a set of pre-defined standard aspect ratios and an option to lock the current one.

> The top-left corner continue to be free dragging,the numeric inputs are unchanged.

I understand that this allows to re-position the rubber band without disengaging the aspect ratio lock, but I don't like the UX, namely:

1. the bottom-right grip jumps away when dragged
2. spinboxes get out of sync with the rubber band
3. swapping checkbox description with a tooltip-style comment resizes the dialog when video width is small

I would love to pull the changes now and provisionally fix these issues myself at the cost of the top-left corner being disabled when the aspect ratio lock checkbox is checked. This will ensure that they make it into the 2.7.7 release.

Alternatively, I would return to this topic right after the release which is already greatly overdue.

Which option do you prefer?

szlldm:
That AR selection menu is a good idea. 
About the UX:
1. i thought about it a lot, but currently i cant see a solution that will keep the apsect ratio and wont jump
2. while tested, didnt see such  
3. agree

If this isn’t too much work for you, let be the first option.
Thank you for the feedback :)

eumagga0x2a:
The UX issues which bothered me should be fixed now. Feedback welcome, especially on possible issues I probably missed.

Will try to look into the new PR soon, thank you.

szlldm:
It looks like it works. Thank you.

PR #200, opened 2021-01-12
MPlayer eq2 video filter luma histogram

szlldm:
Hi!

I took the luma histogram from the Contrast filter, and hacked into the MPlayer eq2 video filter.
Also added a "Show full preview" checkbox. When selected, the effects are shown on the full area of the preview, otherwise only on the right half (same as before this commit).


eumagga0x2a:
Thank you very much.

PR #201, opened 2021-01-13
More bugfixes and improvements (Asharp, MSharpen, MPlayer eq2, swResize)

szlldm:
Hi!

The Asharp filter had bugz, the spinboxes were way out of range, considering the valid range of the filter parameters.
Also have added sliders for convenience (they are in sync with the spinboxes).

The Msharpen filter was also extended with sliders (they are in sync with the spinboxes).
I added a checkbox (& under the hood code) for a "Process chroma" option.

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/asharp/qt4/Q_asharp.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/asharp/qt4/Q_asharp.cpp
@@ -20,6 +20,7 @@
 
 #include "Q_asharp.h"
 #include "ADM_toolkitQt.h"
+#include "math.h"
 
 //
 //	Video is in YV12 Colorspace
@@ -45,7 +46,9 @@
 
 
         connect( ui.horizontalSlider,SIGNAL(valueChanged(int)),this,SLOT(sliderUpdate(int)));
-#define SPINNER(x) connect( ui.doubleSpinBox##x,SIGNAL(valueChanged(double)),this,SLOT(valueChanged(double))); 
+#define SPINNER(x) connect( ui.doubleSpinBox##x,SIGNAL(valueChanged(double)),this,SLOT(valueChanged(double))); \
+		   connect( ui.horizontalSlider##x,SIGNAL(valueChanged(int)),this,SLOT(valueChangedSlider(int)));
+
           SPINNER(Treshold);
           SPINNER(Strength);
           SPINNER(Block);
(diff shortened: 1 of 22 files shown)


szlldm:
MPlayer eq2: bugfix
I have just noticed, that the red and blue gamma sliders operate as them being swapped. It was fine in 2.7.6.
There was a bug in the processing code, the filters for the U and V planes was swapped. But it remained hidden in prev. versions.
As i suspect around commit 88640389c5aef9f32afeb890ace9f0fe5393ce66 there was a corrective U-V plane swapping, and that brought out the problem. 

swResize: improvement
The "Round to the Nearest Multiple of 16" checkbox has been replaced with a ComboBox. There are now multiple choices (none, 16, 8, 4).

eumagga0x2a:
Thank you, it would be great if you could provide isolated fixes for functional bugs you noticed. Anything else will need to wait until after release, with exception of the "Unknown parameter" checkbox, which really needs to get a correct description.

For example, fixing the string "HighQuality" in msharpen.ui invalidates 10 existing translations, which would be justified in other circumstances, but IMHO not now.

I would suggest a workflow of creating a feature branch for each topic and splitting large-scale changes in narrow-scoped commits. This would help a lot to understand the changes later. I would also ask to avoid including merge commits into PR when possible.

For these reasons I cannot pull this big one, unfortunately.

> I have just noticed, that the red and blue gamma sliders operate as them being swapped. It was fine in 2.7.6.

Good catch! The meaning of PLANAR_U and PLANAR_V enumerators was unswapped recently in 3d57249 affecting all code matching anything to these enumerators.

> The "Round to the Nearest Multiple of 16" checkbox has been replaced with a ComboBox. There are now multiple choices (none, 16, 8, 4).

The option to enforce rounding to a multiple of 16 in swResize is dictated by the nature of video compression, smaller values don't make much sense.

szlldm:
Sorry for the inconvenience. I'm very noob with git. So i should create separate branches for every PRs?
The multiple of 16 in swResize does not work at 1080p (i know its probably stored at 1088p). As far as i know modern video encoders, such as x264, support non-mod16 resolutions.

szlldm:
I will look up the proper workflow, to contribute the right way.

eumagga0x2a:
> So i should create separate branches for every PRs?

Yes, please, that would be lovely. If an example helps, please have a look at contributions by xyzzy42.

Multiple descriptions of the standard workflow exist, e.g. https://github.com/susam/gitpr or https://gist.github.com/Chaser324/ce0505fbed06b947d962

> The multiple of 16 in swResize does not work at 1080p (i know its probably stored at 1088p). As far as i know modern video encoders, such as x264, support non-mod16 resolutions.

I acknowledge all that, but this is exactly the reason I don't think we need smaller rounding values. Therefore I am reluctant to accomodate the special case of 1080.

PR #203, opened 2021-01-14
ASharp valid range of spinboxes fixed, added sliders for convenience

szlldm:
The ranges of the spinboxes were 0 .. 99
According to [http://avisynth.nl/index.php/ASharp]:
>Unsharp masking threshold range is from 0 to 32.
>Adaptive sharpening strength must be between 0 and 16.
>Block adaptive sharpening must be less than 4.
I double checked with the underlying code, i can confirm these ranges apply here too.

Added slider controls for the spinboxes for convenience (they are in sync with the spinboxes).

Changed "Unknown flag" to "High quality block filtering"

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/asharp/qt4/Q_asharp.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/asharp/qt4/Q_asharp.cpp
@@ -20,6 +20,7 @@
 
 #include "Q_asharp.h"
 #include "ADM_toolkitQt.h"
+#include "math.h"
 
 //
 //	Video is in YV12 Colorspace
@@ -45,7 +46,9 @@
 
 
         connect( ui.horizontalSlider,SIGNAL(valueChanged(int)),this,SLOT(sliderUpdate(int)));
-#define SPINNER(x) connect( ui.doubleSpinBox##x,SIGNAL(valueChanged(double)),this,SLOT(valueChanged(double))); 
+#define SPINNER(x) connect( ui.doubleSpinBox##x,SIGNAL(valueChanged(double)),this,SLOT(valueChanged(double))); \
+		   connect( ui.horizontalSlider##x,SIGNAL(valueChanged(int)),this,SLOT(valueChangedSlider(int)));
+
           SPINNER(Treshold);
           SPINNER(Strength);
           SPINNER(Block);
@@ -102,6 +105,17 @@ void Ui_asharpWindow::showEvent(QShowEvent *event)
 }
 
 #define MYSPIN(x) w->doubleSpinBox##x
+#define MYSLIDER(x) w->horizontalSlider##x
+
+void Ui_asharpWindow::valueChangedSlider(int f)
+{
+	Ui_asharpDialog *w=(Ui_asharpDialog *)myCrop->_cookie;
+	MYSPIN(Treshold)->setValue((double)MYSLIDER(Treshold)->value() / 100.0);
+	MYSPIN(Strength)->setValue((double)MYSLIDER(Strength)->value() / 100.0);
+	MYSPIN(Block)->setValue((double)MYSLIDER(Block)->value() / 100.0);
+	valueChanged(0);
+}
+
 //************************
 uint8_t flyASharp::upload(void)
 {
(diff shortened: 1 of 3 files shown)


avidemux_plugins/ADM_videoFilters6/asharp/qt4/asharp.ui
```diff
@@ -125,22 +113,70 @@
...
+     </property>
+     <property name="orientation">
+      <enum>Qt::Horizontal</enum>
+     </property>
+    </widget>
+   </item>
+   <item row="1" column="5">
+    <widget class="QLabel" name="label_4">
+     <property name="text">
+      <string>(0 == disabled)</string>
```
eumagga0x2a, line 176: Could we please do without this label (and for optical reasons probably also without the entire fifth column)? I feel it states the obvious. If you feel some hint to be necessary, maybe using setSpecialValueText() to confirm for the user that "0" means "disabled" instead?

szlldm: OK, a will remove it.

avidemux_plugins/ADM_videoFilters6/asharp/qt4/Q_asharp.cpp
```diff
@@ -102,6 +105,17 @@ void Ui_asharpWindow::showEvent(QShowEvent *event)
 }
 
 #define MYSPIN(x) w->doubleSpinBox##x
+#define MYSLIDER(x) w->horizontalSlider##x
+
+void Ui_asharpWindow::valueChangedSlider(int f)
+{
+	Ui_asharpDialog *w=(Ui_asharpDialog *)myCrop->_cookie;
```
eumagga0x2a, line 112: We don't need to go this way as we stay within the class, do we? Is the purpose just to re-use the define?

avidemux_plugins/ADM_videoFilters6/asharp/qt4/asharp.ui
eumagga0x2a: Thanks!

avidemux_plugins/ADM_videoFilters6/asharp/qt4/Q_asharp.cpp
szlldm: IMHO fever or none macro re-define is better solution.
Should i change it to `Ui_asharpDialog *w=&ui;` ?

eumagga0x2a: I meant e.g.

```
void Ui_asharpWindow::valueChangedSlider(int f)
{
#define SYNCSPIN(x) ui.doubleSpinBox##x->setValue(ui.horizontalSlider##x->value() / 100.);
    SYNCSPIN(Treshold)
    SYNCSPIN(Strength)
    SYNCSPIN(Block)

    valueChanged(0);
}
```
Explicit cast is no needed.

eumagga0x2a: My bad, explicit cast here needed, I got lost in brackets.

szlldm: I can rewrite if you want.

eumagga0x2a: I would suggest I pull it now and fix a few remaining UX issues (tab order, single step for spinboxes) myself.

eumagga0x2a:
Thank you very much.

avidemux_plugins/ADM_videoFilters6/asharp/qt4/asharp.ui
eumagga0x2a: If Strength is set to 0, the effect is the same as when it is set to maximum. To disable the filter, it is actually necessary to set it to a small vaue > 0. Do you have an idea what is wrong with the backend code?

`void asharp_run_c(uc* planeptr, int pitch, int height, int width, int T, int D, int B, int B2, bool bf, uint8_t *lineptr);`

indeed gets D set to 0 if param.d is equal zero and 8192 when set to 16 (maximum).

szlldm: According to the avisynth documentation this is the expected behaviour. If > 0 then the unsharp mask applied adaptively, if disabled (== 0) the mask get applied as is.
It would may be the proper way, to add a checkbox, to enable or disable this numeric entry.
Reading this documentation, i came to the conclusion that an another checkbox should be applied on the Block adaptive entry.

PR #204, opened 2021-01-14
MSharpen bugfix and improvements

szlldm:
Hi!

There was a bug in the filter, the "invstrength" variable should be binded to "_param.strength", but sometimes it has not been updated.
Added sliders for convenience (they are in sync with the spinboxes).
Also added a "Process chroma" option (intense sharpening can cause color artifacts, if applied on chroma).
I left the "HighQuality" string as is, to not break translations.
The UI tab order is set.

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/mSharpen/ADM_vidMSharpen.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/mSharpen/ADM_vidMSharpen.cpp
@@ -59,6 +59,7 @@ Msharpen::Msharpen(ADM_coreVideoFilter *in,CONFcouple *couples)
     {
         _param.mask=0;       // Show mask
         _param.highq=1;
+        _param.chroma=0;
         _param.strength=100;	
         _param.threshold=15;	
     }
@@ -96,9 +97,9 @@ Msharpen::~Msharpen(void)
 */
 const char *Msharpen::getConfiguration(void)
 {
-   static char conf[80];
+   static char conf[160];
     conf[0]=0;
-    snprintf(conf,80," Msharpen Strength:%d Threshold:%d",_param.strength,_param.threshold);
+    snprintf(conf,160," Msharpen Strength:%d Threshold:%d Process chroma:%s",_param.strength,_param.threshold,(_param.chroma ? "true":"false"));
     return conf;
 }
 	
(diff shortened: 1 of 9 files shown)


avidemux_plugins/ADM_videoFilters6/mSharpen/ADM_vidMSharpen.cpp
```diff
@@ -118,7 +119,9 @@ ADMImage *src,*blur,*dst;
 
     dst->Pts=src->Pts;
 	
-    for (int i=0;i<3;i++)
+    invstrength=255-_param.strength;	// bugfix
```
eumagga0x2a, line 122: Would you please help me to understand why this is necessary?

In other words, in which circumstances can parameters be modified after object is constructed and getNextFrame() is called? I always thought that the current design tears down the filter chain and creates it anew at output start, but I can easily be in error.

szlldm: The bug can be reproduced the following way:
-load a video, go filters, select msharpen
-enter value (for example 255) to strength, then click OK
-click preview in the Video Filter Manager, and the result will be crap
-now enter the instantiated msharpen filter, and do nothing but click OK
-now the preview will be okay

szlldm: It looks like it only affects the preview of Video Filter Manager.
Is i understand it now, the invstrength only get value at object construction; leaving the configuration UI, the same object stay while in Video Filter Manager.

eumagga0x2a: Yes, there is a bug somewhere in filter chain code. When a filter is already present in the list of active filters, opening and accepting its configuration dialog creates a new instance while the old one is deleted => parameters are refreshed. This doesn't happen when a filter is initiallly added to the list.

eumagga0x2a: Well, the only thing which might be a bug is rebuilding filter chain when only the last filter was modified. You are absolutely right – we must be able to deal with configuration refresh between filter construction and getNextFrame() call.

However, I think it should be enough to do it just once in Msharpen::configure(void)

```
@@ -535,6 +544,7 @@ uint8_t r=0;
     if(DIA_msharpen(copy,this->previousFilter))
     {
         _param=copy;
+        invstrength=(_param.strength < 255)? 255-_param.strength : 0;
         return true;
     }
     return false;
```
rather than at every getNextFrame() invocation.

szlldm:
Sorry something went wrong ....

eumagga0x2a:
Yeah, a stray merge. As all your changes are in one big commit, you could easily just create a new branch from an earlier state, apply the patch, commit it and open a new PR which I would immediately pull (I have all small bits of polish primarily related to avoiding signal cascades from valueChangedSlider() slot ready).

PR #205, opened 2021-01-15
MSharpen corrected bugfix and improvements (sliders & process chroma option)

szlldm:
>There was a bug in the filter, the "invstrength" variable should be binded to "_param.strength", but sometimes it has not been updated.
>Added sliders for convenience (they are in sync with the spinboxes).
>Also added a "Process chroma" option (intense sharpening can cause color artifacts, if applied on chroma).
>I left the "HighQuality" string as is, to not break translations.
>The UI tab order is set.

I hope its okay this way.

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/mSharpen/ADM_vidMSharpen.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/mSharpen/ADM_vidMSharpen.cpp
@@ -59,6 +59,7 @@ Msharpen::Msharpen(ADM_coreVideoFilter *in,CONFcouple *couples)
     {
         _param.mask=0;       // Show mask
         _param.highq=1;
+        _param.chroma=0;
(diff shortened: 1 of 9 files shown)


eumagga0x2a:
Thank you very much.

eumagga0x2a:
Could you please check whether the output with mask enabled is correct when chroma processing is enabled? Additionally, with chroma proc on and HQ off, there is a stripe of colour at the bottom. Is it expected? Can it be fixed?

szlldm:
> Could you please check whether the output with mask enabled is correct when chroma processing is enabled?

It is the same as in previous versions.


> Additionally, with chroma proc on and HQ off, there is a stripe of colour at the bottom. Is it expected? Can it be fixed?

I did noticed too. But it is there in previous versions. It must be with the backend code, currently i dont know if it is fixable. I will look into that sometime.

PR #206, opened 2021-01-15
[asharp/qt4] UI add checkboxes for proper enable/disable option

szlldm:
Backend code behaviour: 
>Adaptive strength is enabled when the D parameter > 0. Disabled if == 0.
>Block adaptive sharpening is disabled ,when the B parameter is negative.

For clarity, checkboxes were added, to handle enabled/disabled states.

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/asharp/ADM_vidAsharp.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/asharp/ADM_vidAsharp.cpp
@@ -82,6 +82,8 @@ ASharp::ASharp(ADM_coreVideoFilter *in,CONFcouple *couples) : ADM_coreVideoFilte
             _param.d=4;
             _param.b=-1;
             _param.bf=false;
+            _param.d_shadow=4;
+            _param.b_shadow=0;
         }
         lineptr=new uint8_t[info.width];
         update();
--- a/avidemux_plugins/ADM_videoFilters6/asharp/ADM_vidAsharp.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/asharp/ADM_vidAsharp.cpp
@@ -82,8 +82,6 @@ ASharp::ASharp(ADM_coreVideoFilter *in,CONFcouple *couples) : ADM_coreVideoFilte
             _param.d=4;
             _param.b=-1;
             _param.bf=false;
-            _param.d_shadow=4;
-            _param.b_shadow=0;
         }
         lineptr=new uint8_t[info.width];
         update();
--- a/avidemux_plugins/ADM_videoFilters6/asharp/ADM_vidAsharp.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/asharp/ADM_vidAsharp.cpp
@@ -82,6 +82,8 @@ ASharp::ASharp(ADM_coreVideoFilter *in,CONFcouple *couples) : ADM_coreVideoFilte
             _param.d=4;
             _param.b=-1;
             _param.bf=false;
+            _param.d_enabled = (_param.d > 0);
+            _param.b_enabled = (_param.b >= 0);
         }
         lineptr=new uint8_t[info.width];
         update();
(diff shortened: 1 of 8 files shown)


eumagga0x2a:
Thanks you, a toggle for block adaptive filtering definitely makes sense as there haven't been a way to disable it in the UI.

The effect of adaptive strength when disabled being identical when set to maximum is utterly confusing IMHO. Are you sure it is not a bug in Avidemux implementation? If not, I am really inclined to trick (i.e. allow 0 as parameter value, but internally translate it to a very small non-zero value depending on toggle state).

I feel that the state of toggles needs to be preserved across sessions, i.e. the set of parameters in asharp struct needs to be extended. I think also that it needs to be decoupled from spinboxes and sliders. It is fine to disable them when the corresponding checkbox is unchecked, but they need to preserve the value between sessions.

szlldm:
> The effect of adaptive strength when disabled being identical when set to maximum
Actually it is not the same. When disabled, and threshold set to max, the macroblock and submacroblock boundaries can be seen (depending on the source), with enabled & max value they not. In the disabled case it is pure unsharp masking. When enbaled and set to max, it will still be adaptive.

> Are you sure it is not a bug in Avidemux implementation?
As far as i can determine, it works same as in avisynth.

The toggle states and the corresponding numeric values are coupled in the current implementation, as the backend algorithm dictates.

>  It is fine to disable them when the corresponding checkbox is unchecked, but they need to preserve the value between sessions.
They are preserved, while configuring the filter (for user's convenience). Applying and closing the configurator, i can not see a reason why sould they be kept.

eumagga0x2a:
> > The effect of adaptive strength when disabled being identical when set to maximum
> 
> Actually it is not the same. When disabled, and threshold set to max, the macroblock and submacroblock boundaries can be seen (depending on the source), with enabled & max value they not. In the disabled case it is pure unsharp masking. When enbaled and set to max, it will still be adaptive.

Indeed, they looked the same only at the default threshold value. Thank you for pointing it out.

> The toggle states and the corresponding numeric values are coupled in the current implementation, as the backend algorithm dictates.

I don't understand why the toggle for adaptive strength is needed then. Disabling by moving slider fully to the left is sufficient then (eventually with special value text added for clarification).

> > It is fine to disable them when the corresponding checkbox is unchecked, but they need to preserve the value between sessions.
> 
> They are preserved, while configuring the filter (for user's convenience). Applying and closing the configurator, i can not see a reason why sould they be kept.

The reason is that the user reopening the configuration dialog will find it in a different state than when closing = bad UX.

szlldm:
> I don't understand why the toggle for adaptive strength is needed then. Disabling by moving slider fully to the left is sufficient then (eventually with special value text added for clarification).

I think its bad UX if you move the slider just a litle bit, and the result is drastically different. The spearate checkbox for enable/disable is more clean.

> The reason is that the user reopening the configuration dialog will find it in a different state than when closing = bad UX.

I will indtroduce shadow parameters then.

eumagga0x2a:
> > I don't understand why the toggle for adaptive strength is needed then. Disabling by moving slider fully to the left is sufficient then (eventually with special value text added for clarification).
> 
> I think its bad UX if you move the slider just a litle bit, and the result is drastically different. The spearate checkbox for enable/disable is more clean.

I agree, the blame falls onto the developers of the original filter implementation, IMHO not suitable 1:1 for a GUI. This is why I would cheat here, keeping the checkbox and allowing zero in the spinbox, which will be translated to a very small non-zero value behind the curtains.

The checkbox will ensure that zero is passed to the algorithm when checked.

> > The reason is that the user reopening the configuration dialog will find it in a different state than when closing = bad UX.
> 
> I will indtroduce shadow parameters then.

That would be lovely, thanks.

eumagga0x2a:
I'm sorry, I've probably caused a profound misunderstanding. I didn't mean really shadow variables, just two more booleans to store the state of checkboxes. The checkboxes determine what we pass to the backend – either the literal value of the float vars (eventually with some cheating for adaptive strength) or a special value to turn a particular feature off.

szlldm:
That would mean messing with the backend code.

eumagga0x2a:
Why? No, not really. Just in update()

eumagga0x2a:
Great, thank you very much.

PR #207, opened 2021-01-15
[resize] more rounding options

szlldm:
I know, from a technical viewpoint only mod16 makes sense, however its inconvenient from the user's viewpoint.
It is on multpile user's wishlist (including mine :) ).


eumagga0x2a:
I think I'll move the combobox into the horizontal layout with spinboxes for width and height to save screen estate.

Taking it, thank you very much.

PR #208, opened 2021-01-16
[security patch] use encrypted channel for update check

szlldm:
Hi!

The current update check use plain http to get the update info, which can be easily hijacked.
Assuming a hijacked connection, the cases i tested:
1) someone can inject malicious URL, that will popup on startup (inattentive/inexperienced users can be deceived)
2) memory overflow, which can lead to unusable/crashed system (swapping, OOM killer, etc...)
For example i fabricated a file with size around 5 GiB. In a minute (i have fiber :) ) the memory usage of avidemux was over 5 GiB.

If the avidemux.org server has limited resources, you can use github as for example: 
https : // raw.githubusercontent.com/mean00/avidemux2/master/update/

### DIFF
--- a/avidemux/qt4/ADM_update/private_inc/ADM_updateImpl.h
+++ b/avidemux/qt4/ADM_update/private_inc/ADM_updateImpl.h
@@ -22,7 +22,7 @@
 #if 0
 #define ADM_UPDATE_SERVER "http://192.168.0.9/"
 #else
-#define ADM_UPDATE_SERVER "http://www.avidemux.org/update/"
+#define ADM_UPDATE_SERVER "https://www.avidemux.org/update/"
 #endif
 
 // Helper defines to construct URL


eumagga0x2a:
I am not associated with avidemux.org and I don't argue with the first point, but if DOS from memory exhaustion is a concern, shouldn't it be caught in `ADMCheckUpdate::execute()` e.g. with something like `manager.head(request)` and proceeding with `manager.get(request)` only if the size is within bounds?

I have no idea whether this is sufficient as protection – a malicious server could probably send different values for HEAD and GET. By all means, introducing an upper bound for `ba.size()` in `ADMCheckUpdate::downloadFinished()` wouldn't harm.

But shouldn't it all be handled by the maintainer of the project, should it?

szlldm:
Other than this simple fix, this is deep water territory for me. However i am sure, at the point the process reaches ADMCheckUpdate::downloadFinished(), the large amount of received data is already in the memory.
IMHO using https and server with valid certificates is the proper and only sane solution.

PR #209, opened 2021-01-17
[crop/qt5] select aspect ratio comboBox for keep aspect ratio

szlldm:
Option for select (other than the source's) aspect ratio for the "Keep aspect ratio" mode.


eumagga0x2a:
Great, thank you very much.

Quickly getting greedy, I think that the aspect ratio menu lacks the most important entry: an option to keep the current AR, which should also be also the default one.

The other thing is that GUI looks a bit disorganized now. I wonder whether autocrop and reset buttons could be moved to the QDialogButtonBox, freeing space for other widgets.

It is probably easier to investigate these aspects separately.

szlldm:
OK, i will look into that

PR #210, opened 2021-01-18
[crop/qt5] fix reset button behaviour, add "current selection" option for aspect ratio list

szlldm:
Reset button behaviour get fixed, when Keep aspect enabled.
Aspect ratio comboBox get "current selection", also this became the default.


eumagga0x2a:
Thank you very much.

PR #211, opened 2021-01-19
[crop/qt5] spinbox bugfix

szlldm:
- spinbox stepsize increased to 2 (because of YV12)
- on change set spinbox maximums, to prevent inversion (bad UX)
- these maximums prevent to set output resolution lower than 2x2 (if one of the dimension is 0, closing filter config will results Avidemux to crash)
- spinboxes get minimum width, to prevent jerkyness

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/crop/qt5/DIA_flyCrop.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/crop/qt5/DIA_flyCrop.cpp
@@ -410,6 +410,10 @@ uint8_t flyCrop::upload(bool redraw, bool toRubber)
     w->spinBoxRight->setValue(right);
     w->spinBoxTop->setValue(top);
     w->spinBoxBottom->setValue(bottom);
+    w->spinBoxLeft->setMaximum(_w-right-2);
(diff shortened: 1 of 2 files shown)


eumagga0x2a:
>     * spinbox stepsize increased to 2 (because of YV12)

I understand that chroma is shifted when values are odd, but please explain why we have no other choice as to regress https://github.com/mean00/avidemux2/commit/69911b15c606a71573a7b93caa047e27403c0a24

>     * on change set spinbox maximums, to prevent inversion (bad UX)

Is this what produces jerkiness? Is it not enough to catch this in download()? (didn't try).

>     * these maximums prevent to set output resolution lower than 2x2 (if one of the dimension is 0, closing filter config will results Avidemux to crash)

Maybe even any dimension < 8 should be rejected if input is larger. I don't know. Hardly any filter was tested with extremely small input dimensions. We should probably start at the editor.

>     * spinboxes get minimum width, to prevent jerkyness

There is no jerkinness with the previous implementation.

szlldm:
Oh, i see. It need some rethinking.

szlldm:
I close this one, to not to clutter with reverts.

PR #212, opened 2021-01-19
[crop/qt5] spinbox bugfix

szlldm:
-on change set spinbox maximums, to prevent inversion (bad UX)
-these maximums prevent to set output resolution lower than 8x8
-spinboxes get minimum width, to prevent jerkiness (introduced by setMaximum)

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/crop/qt5/DIA_flyCrop.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/crop/qt5/DIA_flyCrop.cpp
@@ -485,6 +485,10 @@ Ui_cropDialog *w=(Ui_cropDialog *)_cookie;
                     bottom++;
             }
         }
+        w->spinBoxLeft->setMaximum(_w-right-8);
+        w->spinBoxRight->setMaximum(_w-left-8);
+        w->spinBoxTop->setMaximum(_h-bottom-8);
+        w->spinBoxBottom->setMaximum(_h-top-8);
         rubber->nestedIgnore++;
         rubber->move(_zoom*left+0.49,_zoom*top+0.49);
         rubber->resize(_zoom*bound(left,right,_w)+0.49,_zoom*bound(top,bottom,_h)+0.49);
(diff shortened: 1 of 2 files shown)


eumagga0x2a:
I'm sorry for not responding here earlier.

After my struggle with mplayerDelogo which led to 1b7ebcebfeb24cebad8b002c881654a74eff8946, I fear that Qt doesn't allow setting maximum for a spinbox dynamically while keeping up/down buttons auto-repeat working. Does auto-repeat still work for you in all spinboxes with this change?

I think, minimum width and height should be enforced at parameter validation and other relevant locations in code, also beyond this filter. I just lack a vision how to design this properly at the moment. Last but not least, if a filter may reject input, it must implement a pass-through operation.

szlldm:
>  Does auto-repeat still work for you in all spinboxes 

Well, it doesn't work here either. I have only tested for up/down arrow keys, those works.

ferdnyc:
> After my struggle with mplayerDelogo which led to [1b7ebce](https://github.com/mean00/avidemux2/commit/1b7ebcebfeb24cebad8b002c881654a74eff8946), I fear that Qt doesn't allow setting maximum for a spinbox dynamically while keeping up/down buttons auto-repeat working. Does auto-repeat still work for you in all spinboxes with this change?

I did a proof-of-concept on this, using PySide6. It seems that auto-repeat is only broken if you update the maximum of the **current** spinbox, while it's being changed. As long as you're careful to only update _other_ spinboxes' maxima, but leave the current widget alone, auto-repeat works fine.

And that should be sufficient, because changing any spinbox's value can't affect its **own** maximum value. That _should_ remain static, whatever the current value is set to. So, whenever the code was updating maximum values, it would've always been calling `setMaximum()` on the _current_ widget with the same value it had before. So, the solution is to add logic to... like... not do that. :rofl:  If the `top` value is updating, only call `bottom.setMaximum()`, as that's the only maximum that could possibly change based on the value of `top`. If `left` is changing, only call `right.setMaximum()`. And so on.

ferdnyc:
My proof-of-concept, if you want to try it yourself, is in [ferdnyc/spinbox_test](https://github.com/ferdnyc/spinbox_test). Just clone it and run `python3 spinbox_test.py` from any Python installation/environment that has PySide6 available, and you'll get this dialog:

image

The "Box Totals:" values will be updated based on the spinbox values and maximum values. Auto-repeat on the spinbox arrows works _unless_ "Update all spinboxes" is checked.

szlldm:
Thanks, will look into that.

PR #213, opened 2021-01-19
[crop/qt5] Drag mode

szlldm:
This feature branch is the continuation of 73b5bdd609af2b982c7de691a3f7037119a14d5e (pending Pull Request #212)

Actually editing video, i was still unsatisfied with the crop filter, because repositioning the selection was hard.
Therefore added a drag mode, by replacing "Keep aspect" chechbox with a comboBox, which has:
-Free resize (this is the original mode)
-Free drag
-Keep aspect ratio

However there is a bug i cant figure out (i suspect, it is in the underlying rubber mechanism):
start dragging, the drag point (top-left corner) can not be dragged beyond the pivot point, determined by the bottom-right corner when drag started. The visible rubber band resizes (moves) as it suppose to do, but it acts as if it were a secondary but invisible rubber band, that dont.
Releasing the drag grip, the dragging can be continued, with the new pivot point.

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/crop/ADM_vidCrop.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/crop/ADM_vidCrop.cpp
@@ -74,6 +74,7 @@ CropFilter::CropFilter(ADM_coreVideoFilter *in,CONFcouple *couples) :ADM_coreVid
             configuration.left=0;
             configuration.right=0;
             configuration.rubber_is_hidden=false;
+            configuration.drag=false;
             configuration.keep_aspect=false;
             configuration.ar_select=0;
         }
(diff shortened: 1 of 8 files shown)


eumagga0x2a:
Thank you. A feature to grab the rubber rectangle at any position outside the grips and drag it around would be indeed cool. You could probably subclass ADM_rubberControl and use mouse events. If it works well, it should go into ADM_rubberControl itself.

It would be great to explore it post-release.

-r for any implementation requiring configuration, however.

szlldm:
Yes, it need more work. I already used this feature in editing, and its neat, apart from the bug.
It really acts like the rubber->resize doesnt work until the grip is released. Weird.
I will look into it, if i have free time.

PR #214, opened 2021-01-23
Draggable rubberband

szlldm:
Extended rubber control with drag.
Also adapted all filter that use rubber control (crop, blacken border, mplayer delogo)


eumagga0x2a:
Brilliant. Thank you very much.

eumagga0x2a:
Purely FYI: on macOS (at least on Catalina) with Qt6, the new default version in Homebrew, QEnterEvent is not fired when mouse hovers above the rubber control, therefore the cursor doesn't change to Qt::SizeAllCursor. Probably a bug in Qt, seems to be macOS-only.

PR #215, opened 2021-01-24
[videoFilters] negative

szlldm:
Just a simple color inverter filter, with selectable planes.
(derived from removePlane)


eumagga0x2a:
Nice, thanks. 

`ptr[x] ^= 255;`

is correct for full-range pictures, sure. But shouldn't the range be taken into consideration? The results look very different when the wrong range is used – and the vast majority of videos uses the limited "MPEG" range.

szlldm:
As far as i know this is color inversion by definition. Had checked other implementations (eg. avisynth), it was the same.

eumagga0x2a:
The problem is that with limited range, luma values < 20 are invalid when inverted, i.e. luminance is clipped. Maybe it is safer to expand the range to full first, then invert:

```
if(image->_range == ADM_COL_RANGE_MPEG)
    image->expandColorRange();
```

eumagga0x2a:
I'm pulling it now as the only issue is very minor and can be safely evaluated later.

Thank you.

PR #216, opened 2021-01-27
New video filter category + new filters

szlldm:
Created a new category "Artistic" in the video filter manager . The existing categories and filters are more or less technical in nature.
There are 3 kickoff filter:

- Color effect (eg. sepia, old photo and more...). Ported (& extended) from frei0r framework (GPLv2)
- Vignette (Lens vignetting effect). Ported (& improved) from frei0r framework (GPLv2)
- Pixelize (pixelize with free PAR)


eumagga0x2a:
Great, thank you very much.

Actually, artColorEffect and artVignette would fit into "Color" and artPixelize into "Sharpness". I'm unhappy with the icon which looks blurry (at least very different from others), but I know how hard it is to create a low-res one.

I think that the new category, if it is really necessary, should precede Miscellaneous in the list. This can be safely done after the change has been merged.

szlldm:
I have plans for more filters (porting and also own ideas).
Perhaps these first three could be clearly categorized into the existing options.
However, for example i want to make a VHS filter. It will affect both color and sharpness.

I will try to make the icon better looking.

PR #218, opened 2021-01-28
[artFilter] VHS effect filter

szlldm:
Most authentic at lower resolutions (eg. 576p).


eumagga0x2a:
Thank you, a great effect indeed!

1. I wonder whether it is necessary to printString() "Processed" as there is no side-by-side comparison and therefore the picture must show the preview of output. This costs a little bit of performance, isn't it?
2. There is a usability problem with horizontalSliderUnSync: because of its huge range, it is difficult to use with mouse wheel, you have to scroll and scroll and scroll... and it hardly moves from the spot. It would be great to have this fixed.
3. I would suggest a distinct description for each checkbox (e.g. "No luma delay", "No chroma delay"). My first thought was a copy&paste error until I read the code.
4. It would be also great to declare all the stuff (myFly, canvas, ui) which doesn't necessarily need to be public in Ui_artVHSWindow as private or protected.

eumagga0x2a:
Great, almost there.

I've extended the functionality of ADM_SliderIndicator so that it can be used for horizontalSliderUnSync too, like

```
#define SPINNER(x,y) ui.horizontalSlider##x->setScale(1,y,2); \
        connect( ui.horizontalSlider##x,SIGNAL(valueChanged(int)),this,SLOT(valueChanged(int)));
        SPINNER(LumaBW,100)
        SPINNER(ChromaBW,100)
        SPINNER(UnSync,10)
```
Two small things I've missed last night: in `flyArtVHS::download`, I wonder whether the precision of `MYSPIN(...)->value() / float;` is lost, the integer might need an explicit cast to float here (if I am not mistaken, clang is very strict here, so the feature would be broken on macOS).

The other one is a bounds check missing in `ADMVideoArtVHS::update`, it is easy to make Avidemux crash in ArtVHSProcess_C with input values it doesn't like.

eumagga0x2a:
Thank you very much.

PR #219, opened 2021-01-28
[videoFilterManager] New icons

szlldm:
There is a better quality icon for OpenGL category in the first commit (07577c1).

The second commit (4c466cc) is just a proposal for the Noise category.


eumagga0x2a:
I apologize for not responding earlier.

The OpenGL icon is definitely an improvement over the current one, but I fear than the idea to use the official logo as the base for icon simply doesn't work with this one.

Regarding the Noise category icon, I am not sure it is any clearer than the current one while the current one has that diagonal main structure which matches both the icon for Scharpness and the icon for Artistic. I cannot embrace the updated version, sorry for that.

szlldm:
OK. I close this PR.

PR #220, opened 2021-01-28
[artFilter] Charcoal/Chalkboard effect filter

szlldm:
Ported from MLT framework (LGPL/GPL).


eumagga0x2a:
Thank you very much for proactively fixing issues.

I wonder whether the algorithm could be improved performance-wise? This filter is super-slow on my hardware, not reaching real-time even with 720p at 25fps. Maybe is multi-threading possible? I don't know.

I need to draw a line, else the release will never happen. I'll come back to this one once it is done.

PS + OT: If it might interest you to go straight for the holy grail, Avidemux reaches > 500 fps decoding H.264 at 720p in software with multi-threading enabled (2 cores i3 with HT). With VA-API hardware accelerated enabled, the speed drops to ~90 fps. `nv12_to_uv_mmx` in avidemux_core/ADM_coreImage/src/ADM_imageUtils.cpp:248 is the bottleneck.

Don't even try the C version, it is at least twice as slow.

szlldm:
Its fine. Until then i will try to optimize the code more. (The current code is already at least 2x faster than it was right after the initial porting)

So VA-API is a hw decelerator? :)

eumagga0x2a:
> The current code is already at least 2x faster than it was right after the initial porting

Got it, thanks. I really hope that it can be made at least twice as fast as it is now.

> So VA-API is a hw decelerator? :)

As long as hw surfaces don't need to be converted to regular images with deinterleaved chroma, VA-API is great. Even going the oppositve way (YV12 --> NV12) is pretty fast. But NV12 --> YV12 is an unmitigated disaster.

The same problem affects DXVA2 hw accel on Windows which also requires NV12 --> YV12 conversion and additionally cannot short-circuit hw decoder and video display.

szlldm:
Currently i could only test it on an old Core 2 Quad, but 2x faster (saving with null encoder).

szlldm:
OT:
ADM_imageUtils_asm.asm / nv12_to_u_v_one_line
If i'm not mistaken
       psllw          m0,8
       psrlw          m0,8
       psllw          m1,8
       psrlw          m1,8
could be replaced with
       pand m0,m4
       pand m1,m4
of course m4 must be initialized with the right value, before .nv12_again

Maybe it worth a try, to move
       movq           [dstvq],m0
just after 
       packuswb       m0,m1
(its efficacy depends on how deep&smart the out-of-order mechanism is)

eumagga0x2a:
> Currently i could only test it on an old Core 2 Quad, but 2x faster (saving with null encoder).

Much better, thanks. 720p@25 runs in real time now on my hw, 1080p@25 is still off-limits.

I'm merging this PR. It would be great if you could come back to this filter to explore further speed-ups during the next release cycle, however.

PR #221, opened 2021-01-30
[coreImage] nv12_to_uv_mmx

szlldm:
I did code this blind ,  cant test it.
It should work (but not guaranteed).
It may or may not improve runtime.
¯\_(ツ)_/¯

### DIFF
--- a/avidemux_core/ADM_coreImage/src/ADM_imageUtils_asm.asm
+++ b/avidemux_core/ADM_coreImage/src/ADM_imageUtils_asm.asm
@@ -105,28 +105,26 @@ cglobal YUV444_chroma, 4,4,8, src, dst, dst2, w4
 ;
 INIT_MMX mmx
 cglobal nv12_to_u_v_one_line,4,4,4, w8, dstu, dstv, src
-
+       pxor           m4, m4
+       pcmpeqb        m4, m4
+       psllw          m4,8
+       psrlw          m4,8
 .nv12_again
        movq           m0 , [srcq]
        movq           m1 , 8[srcq]
        movq           m2, m0
        movq           m3, m1
 
-       psllw          m0,8
-       psrlw          m0,8
-
-       psllw          m1,8
-       psrlw          m1,8
-
+       pand           m0, m4
+       pand           m1, m4
 
        packuswb       m0,m1
+       movq           [dstvq],m0
 
        psrlw          m2,8
        psrlw          m3,8
 
        packuswb       m2,m3
-
-       movq           [dstvq],m0
        movq           [dstuq],m2
 
        add            srcq,16


eumagga0x2a:
Thank you, amazing, but the code actually works, albeit it doesn't speed up things, unfortunately.

(I tried it first by using the dummy video filter modified to do a roundtrip YV12 --> NV12 --> YV12)

Here I owe you an apology as I forgot that I had changed `admLibVA::surfaceToAdmImage` myself to do the NV12 --> YV12 conversion not by `convertFromNV12()` I pointed to but by libswscale. The slow operation reflects the performance of libswscale while `convertFromNV12()` is even much slower on my hardware (but not on an i7 as I heard), I get hardly 60fps for 720p source when using the internal function. Unfortunately, the patch doesn't make it faster.

What puzzles me, my old main PC was more than 2x faster in the roundtrip test with the same 720p source than the i3 machine with `admLibVA::surfaceToAdmImage` rewired to `convertFromNV12()`.

Skipping download from hw surface in dummy filter by

```
diff --git a/avidemux_plugins/ADM_videoFilters6/dummy/dummyVideoFilter.cpp b/avidemux_plugins/ADM_videoFilters6/dummy/dummyVideoFilter.cpp
index 261b2b8d2..c580408e2 100644
--- a/avidemux_plugins/ADM_videoFilters6/dummy/dummyVideoFilter.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/dummy/dummyVideoFilter.cpp
@@ -33,6 +33,8 @@ public:
         virtual bool         getCoupledConf(CONFcouple **couples) ;   /// Return the current filter configuration
 		virtual void setCoupledConf(CONFcouple *couples);
         virtual bool         configure(void) {return true;}             /// Start graphical user interface
+private:
+                ADMImage     *scratch;
 };
 
 // Add the hook to make it valid plugin
@@ -56,6 +58,10 @@ UNUSED_ARG(setup);
 
     // By default the info field contains the output of previous filter
     // Tweak it here if you change fps, duration, width,...
+    scratch = new ADMImageDefault(in->getInfo()->width, in->getInfo()->height);
+    memset(scratch->_planes[0],179,scratch->_planeStride[0]*in->getInfo()->height);
+    memset(scratch->_planes[1],120,scratch->_planeStride[1]*in->getInfo()->height/2);
+    memset(scratch->_planes[2],57,scratch->_planeStride[2]*in->getInfo()->height/2);
 }
 /**
     \fn dummyVideoFilter
@@ -71,8 +77,13 @@ dummyVideoFilter::~dummyVideoFilter()
 */
 bool dummyVideoFilter::getNextFrame(uint32_t *fn,ADMImage *image)
 {
+    ADMImageRefWrittable ref(previousFilter->getInfo()->width, previousFilter->getInfo()->height);
     // since we do nothing, just get the output of previous filter
-    return previousFilter->getNextFrame(fn,image);
+    if(false == previousFilter->getNextFrameAs(ADM_HW_LIBVA,fn,&ref))
+        return false;
+    image->duplicate(scratch);
+    image->copyInfo(&ref);
+    return true;
 }
 /**
     \fn getCoupledConf
```
just to benchmark the performance of the hw decoder gives me > 500 fps with VA-API on the i3 machine.

szlldm:
Maybe SSE could improve, as it has double width registers.
So far i only used SSE with intrinsics, dont know how crossplatform compatible it is.

PR #223, opened 2021-02-01
[artCharcoal] Optimize more

szlldm:
+40% fps
on my old i5 it can reach  80fps@1080p (i think it sufficient, as few useful encoder can keep up with this frame rate )


eumagga0x2a:
On my i5 laptop, it topped just above 60fps@1080p with multi-threading in libavcodec enabled, which is a vast improvement, thank you very much.

PR #224, opened 2021-02-01
[artVHS] improve sync error by adding filtering option

szlldm:
new slider control for filtering sync error
more authentic look


eumagga0x2a:
Thank you.

PR #225, opened 2021-02-03
 [videoFilter] New filter: Color temperature

szlldm:


eumagga0x2a:
Thank you very much.

eumagga0x2a:
I understand that I am at fault here, but could we drop the call to expandColorRange() in `ADMVideoColorTemp::ColorTempProcess_C`? It has a huge performance penalty (~halves fps at 720p) and clipping happens this way or another, depending on temperature and angle.

I could imagine scaling temperature proportional to color range instead:

```
diff --git a/avidemux_plugins/ADM_videoFilters6/colorTemp/ADM_vidColorTemp.cpp b/avidemux_plugins/ADM_videoFilters6/colorTemp/ADM_vidColorTemp.cpp
index 35487b87e..8492e0c18 100644
--- a/avidemux_plugins/ADM_videoFilters6/colorTemp/ADM_vidColorTemp.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/colorTemp/ADM_vidColorTemp.cpp
@@ -46,6 +46,8 @@ DECLARE_VIDEO_FILTER_PARTIALIZABLE(   ADMVideoColorTemp,   // Class
 */
 void ADMVideoColorTemp::ColorTempProcess_C(ADMImage *img, float temperature, float angle)
 {
+    static const float scale = (240.-16.)/255.;
+
     int width=img->GetWidth(PLANAR_Y); 
     int height=img->GetHeight(PLANAR_Y);
     int ystride,ustride,vstride;
@@ -60,7 +62,10 @@ void ADMVideoColorTemp::ColorTempProcess_C(ADMImage *img, float temperature, flo
     int ushift, vshift;
 
     if(img->_range == ADM_COL_RANGE_MPEG)
-        img->expandColorRange();
+    {
+        ushiftf *= scale;
+        vshiftf *= scale;
+    }
 
     // Y plane unchanged
 
```
(by the way, very contra-intuitively, optimizing by avoiding repeated computation of std::sin and std::cos doesn't have any positive effect on my hardware)

PR #227, opened 2021-02-04
[videoFilter] New filter: Blur

szlldm:
Almost Gaussian blur with adjustable radius.
Based on Mario Klingemann's stack blur algorithm.
Well it is slow. I tried to optimize it, a get around +10%. (On my old i5 i get 50fps@720p)


szlldm:
Added a Box blur mode (it is 50% faster than the "Near Gauss").
Also added a 2 pass Gaussian blur (obviously it is twice as slow as "Near Gauss", but has a slightly better quality).

butterw:
- In the preview ui it would be useful to able to have a switch to toggle the effect on/off quickly to be able to see the difference. This would probably be true also for other effects (especially sharpeners on low strength).

- For blur and pixelize filter a feature would be to only apply the effect to a selected-region. It's probably not all that useful in practice though (unless you also have face detection).

szlldm:
> In the preview ui it would be useful to able to have a switch to toggle the effect on/off quickly to be able to see the difference. 
This would probably be true also for other effects (especially sharpeners on low strength).

Yes, it would be nice, however it should be implemented globally (i.e. it should work the same in every filter's preview), not per filter. For example clicking on the preview would bring the original, then releasing the click would bring back the filtered preview.

butterw:
Yes it should be defined globally. The feature would then be available on the filters with preview and the global preview. 
To toggle state, clicking on the preview or displaying a switch in the preview playback toolbar would work.

eumagga0x2a:
Thank you very much, merged. All refinements can be safely done later in the repository.

> To toggle state, clicking on the preview or displaying a switch in the preview playback toolbar would work.

The former would break at least the logo filter and could conflict with rubber band, I also doubt that it could be made keyboard-accessible. The latter is the way to go.

Replacing button labels in flyDialog control with icons would allow to reduce the screen real estate occupied by buttons, freeing space for a toggle. Having a preview vs passthrough toggle in flyDialog by default would also allow to drop side-by-side preview in many filters.

szlldm:
Well, first in the Wavelet Sharpener i placed a "Peek Original" button, which only shows the input frame until the button is pressed.
It is part of the filter dialog, not flyDialog, because i realized a global button would break some filter's preview...
I think it has the best UX so far.

PR #230, opened 2021-02-09
[videoFilters] Quick fixes

szlldm:
[negative]: extend color range if the source is ADM_COL_RANGE_MPEG
Other filters: copy the whole image (incl. infos&flags!) to the output (instead just the planes) in the preview processor.


eumagga0x2a:
Thanks!

PR #231, opened 2021-02-10
[videoFilter] New filter: Chroma key

szlldm:
Replace "green screen" with an image.
The loaded picture will be scaled to the video.


eumagga0x2a:
Great, thanks!

PR #235, opened 2021-02-17
[videoFilter] New filter: Analyzer

szlldm:
This is a null filter, it do nothing, but in the filter dialog there are:

- Vectorscope
- Waveform scopes (RGB & YUV)
- Hystograms (RGB & YUV)

There is a small bug i can not rid off; opening the filter dialog, the video preview wont fit, until the first manual dialog resize.
All i know is, that the canvas report false width&height.
I even moved the fly dialog constructor call into the showEvent handler, but the bug still persist.


eumagga0x2a:
> There is a small bug i can not rid off; opening the filter dialog, the video preview wont fit, until the first manual dialog resize.

Actually, the size of all QGraphicsView widgets is completely wrong unless the dialog is resized manually, not just the canvas. The problem disappears when scale(0.5, 0.5) calls are removed. Unsure what is going on here either.

There is also a tiny trivial bug which upsets the compiler – flyAnalyzer::download(void) and flyAnalyzer::upload(void) don't return anything.

szlldm:
> The problem disappears when scale(0.5, 0.5) calls are removed.

Yes its true, but the minimum size of the dialog is too big, if not scaled. Its fine on my 1080p display, but on smaller screens wouldnt fit.

szlldm:
I think i have solved/hid the problem, at least visually.

eumagga0x2a:
> I think i have solved/hid the problem, at least visually.

Unfortunately, this change means that when a user resizes the dialog manually, switches to a different workspace and then returns back, the dialog is automatically resized back to 900x650 as a show event is fired each time the window is minimized and then restored.

I'm going to pull this PR as-is and fix UI/UX in-tree. Luckily, there is no time pressure anymore :-)

szlldm:
Its trivial to call resize once.
However its still bother me, that this is a sweep under the carpet solution :)
I have checked many forums and and tutorials, and did stumble into similar problems, but never the solution...

eumagga0x2a:
Almost everything was in place, just a call to adjustSize() at the end of the first showEvent() was missing.

I moved flyAnalyzer construction back into the dialog constructor to simplify things and cleaned up a bit.

szlldm:
Amazing! Thanks!

PR #236, opened 2021-02-18
[colorTemp] fix label, fix for limited range

szlldm:
Changed "U-V plane angle of attack" label to "Chroma shift angle".
Also modified the algorithm, to better handle the reduced range (16..235) luma.


eumagga0x2a:
Thank you.

PR #237, opened 2021-02-18
[artPixelize] change description

szlldm:


eumagga0x2a:
Please don't be surprised, after merge, I'll add a period to the description to match all the others.

An important question: would you like to be mentioned by nick or alternatively by the full name, if you wish so and provide it, in the release changelog?

szlldm:
Ok.
If You think i deserve to be mentioned, by nick is fine. Thanks.

PR #238, opened 2021-02-18
[vidPartial] Show embedded filter name

szlldm:
Show the partialized filter name in the partial entry of the active filter list.

### DIFF
--- a/avidemux/common/ADM_videoFilter2/src/ADM_vidPartial.cpp
+++ b/avidemux/common/ADM_videoFilter2/src/ADM_vidPartial.cpp
@@ -25,7 +25,7 @@
 extern ADM_coreVideoFilter *ADM_vf_createFromTag(uint32_t tag, ADM_coreVideoFilter *last, CONFcouple *couples);
 extern uint32_t    ADM_vf_getTagFromInternalName(const char *name);
 
-ADM_coreVideoFilter *createPartialFilter(const char *internalName,const char *displayName,CONFcouple *couples);
+ADM_coreVideoFilter *createPartialFilter(const char *internalName,CONFcouple *couples);
 /**
     \class partialFilter
 */
@@ -108,7 +108,7 @@ partialFilter::partialFilter(  ADM_coreVideoFilter *in,CONFcouple *setup) : ADM_
     // Step 1 : Load configuration
     if(!setup)
         ADM_assert(0);
-    // Only keep 4 parameters
+    // Only keep 3 parameters
     char *filterName=NULL;
     if(!setup->readAsString ("filterName",&filterName))
       {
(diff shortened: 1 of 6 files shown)


eumagga0x2a:
Thank you, is the purpose of the patch to get translated description of the partialized filter in the output of `getConfiguration()` instead of untranslated and more technical filter name which is also usually duplicated in the partialized filter configuration printed below?

Why not something like

```
diff --git a/avidemux/common/ADM_videoFilter2/src/ADM_vidPartial.cpp b/avidemux/common/ADM_videoFilter2/src/ADM_vidPartial.cpp
index 3443db602..09f254608 100644
--- a/avidemux/common/ADM_videoFilter2/src/ADM_vidPartial.cpp
+++ b/avidemux/common/ADM_videoFilter2/src/ADM_vidPartial.cpp
@@ -15,6 +15,8 @@
 
 #include "ADM_default.h"
 #include "ADM_coreVideoFilter.h"
+#include "ADM_coreVideoFilterFunc.h"
+#include "ADM_videoFilterApi.h"
 #include "DIA_factory.h"
 #include "DIA_coreToolkit.h"
 #include "ADM_vidMisc.h"
@@ -22,9 +24,6 @@
 #include "partial_desc.cpp"
 #include "avi_vars.h"
 
-extern ADM_coreVideoFilter *ADM_vf_createFromTag(uint32_t tag, ADM_coreVideoFilter *last, CONFcouple *couples);
-extern uint32_t    ADM_vf_getTagFromInternalName(const char *name);
-
 ADM_coreVideoFilter *createPartialFilter(const char *internalName,CONFcouple *couples);
 /**
     \class partialFilter
@@ -312,9 +311,12 @@ void partialFilter::setCoupledConf(CONFcouple *couples)
 */
 const char *partialFilter::getConfiguration(void)
 {
-  sprintf(description,"Partial : %s -- ",ADM_us2plain((uint64_t)(configuration.startBlack)*1000));
+  uint32_t id = ADM_vf_getTagFromInternalName(configuration.filterName.c_str());
+  sprintf(description,"%s: ",ADM_vf_getDisplayNameFromTag(id));
+  strcat(description,ADM_us2plain((uint64_t)(configuration.startBlack)*1000));
+  strcat(description," -- ");
   strcat(description,ADM_us2plain((uint64_t)(configuration.endBlack)*1000));
-  strcat(description," ");
+  strcat(description,"\n");
   strcat(description,sonFilter->getConfiguration());
   return description;
 }
```
in this case? The thing is that we already have filterName string in the configuration, which already uniquely identifies the partialized filter. Having a configuration parameter just as a buffer to hold a value entirely determined by another configuration parameter feels somewhat redundant.

szlldm:
Indeed it is simpler this way :)
The purpose is to show the embedded filter name in a uniform way.
Some filter have their  names in the getConfiguration() (imho these should be omitted , because redundant, and not translated), some dont.

eumagga0x2a:
Thx, I like the idea, but I'm going to commit and push my own patch myself :-))

PR #240, opened 2021-02-19
[ADM_coreVideoCodec] enable VP9 multithreading

szlldm:
Experimental!
The _ADM_ffmp43_ multihreading does not cover VP9 encoded videos.
I can play now 4k60fps VP9 videos buttery smooth with this patch. With single thread it was p.i.t.a.
As far as i know _ADM_codecFFsimple_ / class _decoderFFSimple_ is not used elsewhere, but _ADM_codecFFVP9_  (is this true?).
Im not sure this is the correct way, to enable MT for VP9, but works for now..

### DIFF
--- a/avidemux_core/ADM_coreVideoCodec/src/ADM_codecFFsimple.cpp
+++ b/avidemux_core/ADM_coreVideoCodec/src/ADM_codecFFsimple.cpp
@@ -33,7 +33,7 @@ decoderFFSimple::decoderFFSimple (uint32_t w, uint32_t h,uint32_t fcc, uint32_t 
         return;
 
     AVCodecID id=c->codecId;
-    AVCodec *codec=avcodec_find_decoder(id);
+    codec=avcodec_find_decoder(id);
     if(!codec)
     {
         GUI_Error_HIG(QT_TRANSLATE_NOOP("adm","Codec"),QT_TRANSLATE_NOOP("adm","Internal error finding codec 0x%x"),fcc);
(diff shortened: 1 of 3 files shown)


avidemux_core/ADM_coreVideoCodec/src/ADM_codecFFsimple.cpp
```diff
@@ -56,8 +56,15 @@ decoderFFSimple::decoderFFSimple (uint32_t w, uint32_t h,uint32_t fcc, uint32_t
...
-    printf("[decoderFFSimple] context allocated with thread_count = %d\n",_context->thread_count);
+
     decoderMultiThread();
+    if (_usingMT) {
+        if(codecId==AV_CODEC_ID_VP9) { //TODO investigate what other codecs can be multithreaded!
+            _context->thread_count = _threads;
+            _context->thread_type = FF_THREAD_SLICE;    // this is important! the default FF_THREAD_FRAME wont work with Avidemux
+        }
+    }
+    printf("[decoderFFSimple] context allocated with thread_count = %d\n",_context->thread_count);
```
eumagga0x2a, line 67: This debug message is misleading here as we might have modified thread_count in AVCodecContext.

eumagga0x2a (commented):
I would love to avoid codec-specific quirks in the catch-all decoder. Maybe adding an `applyQuirks()` function (I'm bad at inventing names) to decoderFF which derived classes could reimplement to do stuff right before `avcodec_open2()`?

szlldm:
My main goal was to check multithreading works in VP9. I know it should have been made with capability check, but i was too lazy, as i intended this PR as discussion rather an actual patch :)
The frame based multithreading is clearly not working (maybe the codec provides the decoded frames out of order? idk, just guessing)

> Maybe adding an applyQuirks() function (I'm bad at inventing names) to decoderFF which derived classes could reimplement to do stuff right before avcodec_open2()?

The parent constructor can not call a virtual function from child class :(

eumagga0x2a:
> The parent constructor can not call a virtual function from child class :(

No, and also unnecessary, was a stupid late night idea. Just split the avcodec_open2 part away from the decoderFFSimple ctor in something like decoderFFSimple::finish(void), so that admCreateFFSimple would call it instead of initialized(). Derived classes like decoderFFVP9 will do their stuff with _context first, then call finish().

eumagga0x2a:
Great, thanks!

I'll pull it and make just two small adjustments (finish() should return boolean, so that the later call of initialized() can be dropped and checking capabilities won't harm).

szlldm:
Ok, thanks!
Looking further, why fcc displayed/printed as hexadecimal numbers? It could be displayed as 4 characters, thus would be more meaningful.
_Internal error finding codec 0x31307661_  vs.  _Internal error finding codec "av01"_

eumagga0x2a:
Probably for safety as the failure to find decoder arises mostly when some rubbish is passed as fcc, there are no checks in `fourCC::tostring`.

PR #242, opened 2021-02-22
[videoFilter] New filter: Fit to size

szlldm:
Resize and pad to the specified size.
The padding can be black bars, or echo.


eumagga0x2a:
Awesome, thanks!

Methinks, both addBorder and blackenBorder filters could borrow from the echo option in this one. If only the word "blacken" weren't there...

PR #243, opened 2021-02-23
[Q_mainfilter] fix crash

szlldm:
Closing the video filter manager's preview, the flydialog was not deleted, and that can caused crashes.
Video filter preview crash can be replicated:
    - instantiate any two filter, that has previews in the configuration dialog
    - open the global preview in the video filter manager, click play, then close with OK button
    - open the first filter, and click play, then close with OK button

Crash Dump for Crash
Segfault
 at line 0, file ??
ADM_backTrack

ADM_flyDialog::nextImageInternal()
ADM_flyDialog::nextImage()
ADM_flyDialog::timeout()
QMetaObject::activate(QObject*, int, int, void**)
QTimer::timeout(QTimer::QPrivateSignal)
QTimer::timerEvent(QTimerEvent*)
QObject::event(QEvent*)
QApplicationPrivate::notify_helper(QObject*, QEvent*)
QApplication::notify(QObject*, QEvent*)
QCoreApplication::notifyInternal2(QObject*, QEvent*)
QTimerInfoList::activateTimers()

### DIFF
--- a/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_mainfilter.cpp
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_mainfilter.cpp
@@ -288,7 +288,11 @@ void filtermainWindow::preview(bool b)
 void filtermainWindow::closePreview()
 {
     if (previewDialog)
+    {
         qtUnregisterDialog(previewDialog);
+        delete previewDialog;
+        previewDialog = NULL;
+    }
 }
 
 /**


eumagga0x2a:
Thank you for reporting this crash. Unfortunately, the suggested fix makes the crash even worse on macOS at least (Avidemux segfaults hard right after Step 2).

It looks like several bugs are involved here. For sure, filter seeking gets out of sync, but not just that.

szlldm:
Ok, then i close this PR.
Other occurence i have found is when resize involved in the filter chain, and the closed (but not deleted) preview had different width/height registered, then the actual  w/h changed because of the resizer.

eumagga0x2a:
I don't think the patch is wrong, it covers just a part of the picture.

eumagga0x2a:
Could you please check whether

```
diff --git a/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_mainfilter.cpp b/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_mainfilter.cpp
index 68bf3231c..f29e88fac 100644
--- a/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_mainfilter.cpp
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_mainfilter.cpp
@@ -288,7 +288,14 @@ void filtermainWindow::preview(bool b)
 void filtermainWindow::closePreview()
 {
     if (previewDialog)
+    {
+        if(previewDialog->seekablePreview)
+        {
+            previewDialog->seekablePreview->play(false);
+            previewDialog->seekablePreview->goToTime(0);
+        }
         qtUnregisterDialog(previewDialog);
+    }
 }
 
 /**
```
fixes the crash for you?

Deleting previewDialog in right in `filtermainWindow::closePreview` instead of later during the next call of `filtermainWindow::preview` works great on Linux but results almost always in a segfault in Qt on macOS (like a race condition somewhere in QtCore).

> when resize involved in the filter chain, and the closed (but not deleted) preview had different width/height registered, then the actual w/h changed because of the resizer.

I fear I don't quite follow you here. Could you please elaborate?

szlldm:
>> when resize involved in the filter chain, and the closed (but not deleted) preview had different width/height registered, then the actual w/h changed because of the resizer.

>I fear I don't quite follow you here. Could you please elaborate?

Sorry :)  So, if there is a resizer instantiated amongst the filters, and (after the "closed preview") it is moved up/down, or the resized dimensions are changed, it will lead to crash (and i suspected it happens because the preview want to read a different sized ADM_Image, that is actually provided to it.)

--
It looks like this patch solved the issue. Thank You!

eumagga0x2a:
>  So, if there is a resizer instantiated amongst the filters, and (after the "closed preview") it is moved up/down, or the resized dimensions are changed, it will lead to crash (and i suspected it happens because the preview want to read a different sized ADM_Image, that is actually provided to it.)

This would definitely happen if we reuse the previewDialog. No crash for this reason as we don't.

I've pushed the fix. Thank you very much again for noticing the problem.

PR #244, opened 2021-02-24
[x264] Mezzanine preset

szlldm:
Ultrafast, intra frame only preset for intermediate editing stage encoding.


eumagga0x2a:
On a quick test, ~ a half of bitrate of a true lossless codec (Ut Video) for yuv420p at a comparable encoding speed + can be decoded in hardware.

Thanks!

PR #245, opened 2021-02-25
[videoFilter] New filter: Wavelet Sharpener

szlldm:
Based on Marco Rossini's sharpener. (Heavily optimized, i got 3-4x speedup)
Also there is a new concept for filters: there is a "Peek Original" button. When it is pressed, it will show the unfiltered input, but on release the filtered preview is back again.


eumagga0x2a:
Thank you very much for porting this filter.

I wonder whether the peek button could be moved into ADM_flyDialog to reduce code duplication. I like the concept very much.

Not a big deal for now, but I dislike clarifying printString() overlays for 5 reasons: the first three are performance impact, the last two are the font size being inversely proportional to video resolution due to zooming (almost unreadable at 4K, huge and ugly at low resolutions) and the displayed text being not translatable because the font doesn't include non-ASCII characters (well, kanji or hangeul at this pixel size won't be readable anyway).

szlldm:
I think moving the peek button into ADM_flyDialog would require a separate subclass, because there are filters (for example Crop and Blacken border) where a peek button wouldnt make sense. Also in Color balance there is a Peek ranges button, which would be impossible with this soultion. But its doable.

eumagga0x2a:
A subclass or an additional parameter to addControl() with the default value not to add the peek button.

> Also in Color balance there is a Peek ranges button, which would be impossible with this soultion.

I can be easily wrong here, but why should this become impossible? It should be possible to add the peek ranges button to horizontal layout programmatically, just not via the .ui file.

szlldm:
I meant it is over my skill level :D

PR #246, opened 2021-03-01
[videoFilter] New filter: Grid

szlldm:
Video wall effect.


eumagga0x2a:
A great filter, thanks!

There are a few bugs with the roll option enabled (pts are invalid, output not cleared after seek). I'll fix them, should be easy.

szlldm:
Ok, thanks!

PR #250, opened 2021-03-16
[videoFilter] New filter: DelogoHQ

szlldm:
Clean up arbitrary shaped logo.
Didnt find code/algo for porting, so i wrote one.
Compared to Delogo2:
delogohq


eumagga0x2a:
Brilliant, thank you.

I wonder whether it would make sense to compact groupBoxMask to just a single row or a horizontal layout and move all instructions to a dedicated help dialog or to a tooltip (the former is probably better discoverable). I failed to understand the instructions in their current compressed form and had to read the code to realize that instead of creating a black and white mask (i.e. adding transparency), I needed to create a black and white image which would be used by the filter as a mask. Once I mastered this intellectual challenge, I wished to be able to use the screen real estate more efficiently as I didn't feel like I'd ever need the instructions.

szlldm:
Ok, i will try to do that.

PR #253, opened 2021-03-19
Flydialog Peek Original button + applied to filters

szlldm:
Optional through ADM_flyDialog::addControl(QHBoxLayout *layout, bool addPeekOriginalButton = false);

Applied to most of the art filters, chromaShift, colorTemp, blur and DelogoHQ.
In waveletSharp and waveletDenoise the built in Peek Original buttons were replaced with this.

In contrast, eq2, hue, asharp and mSharpen the old and diverse preview modes were deleted, and Peek Original buttons were added.


eumagga0x2a:
Wonderful, thank you.

PR #255, opened 2021-03-20
[eq2] bugfix

szlldm:
Fix bug introduced in commit 57a5b6b.
(Red and blue sliders were swapped).


eumagga0x2a:
Indeed, the `(ADM_PLANE)i; i++;` part (old habit) was the trigger / culprint. Thanks!

PR #256, opened 2021-03-21
[Video filter manager] improve UX

szlldm:
Deleting a filter, dont jump to the end of the list, but stay at the same position if possible
Show shortcut keys in context menu
Add shortcuts for Move Up/Down, and Make partial


eumagga0x2a:
Now it is possible to avoid using context menu in the first place, which is really good, thank you very much.

PR #257, opened 2021-03-22
[filtermanager] Option for enable/disable instantiated filters

szlldm:
Tested many scenarios, it looks like it works.
What i dont know is, the scriptEngines.

### DIFF
--- a/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_seekablePreview.h
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_seekablePreview.h
@@ -34,7 +34,7 @@ public:
 	Ui_seekablePreviewDialog ui;
         
 public:        
-                Ui_seekablePreviewWindow(QWidget *parent, ADM_coreVideoFilter *videoStream, uint32_t defaultFrame = 0);
+                Ui_seekablePreviewWindow(QWidget *parent, ADM_coreVideoFilter *videoStream, QString const& title=QString(""), uint32_t defaultFrame = 0);
                 ~Ui_seekablePreviewWindow();
 	void    resetVideoStream(ADM_coreVideoFilter *videoStream);
 	uint32_t frameIndex();
--- a/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_seekablePreview.h
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_seekablePreview.h
@@ -34,7 +34,7 @@ public:
 	Ui_seekablePreviewDialog ui;
         
 public:        
-                Ui_seekablePreviewWindow(QWidget *parent, ADM_coreVideoFilter *videoStream, QString const& title=QString(""), uint32_t defaultFrame = 0);
+                Ui_seekablePreviewWindow(QWidget *parent, ADM_coreVideoFilter *videoStream, uint32_t defaultFrame = 0);
                 ~Ui_seekablePreviewWindow();
 	void    resetVideoStream(ADM_coreVideoFilter *videoStream);
 	uint32_t frameIndex();
--- a/avidemux/common/ADM_videoFilter2/include/ADM_videoFilters.h
+++ b/avidemux/common/ADM_videoFilter2/include/ADM_videoFilters.h
@@ -21,6 +21,7 @@
 // Get list
 uint32_t                ADM_vf_getSize(void);
 ADM_coreVideoFilter     *ADM_vf_getInstance(int index);
+bool                    ADM_vf_getEnabled(int index);
 uint32_t                ADM_vf_getTag(int index);
 bool                    ADM_vf_getConfigurationFromIndex(int index,CONFcouple **c);
 
--- a/avidemux/common/ADM_videoFilter2/src/ADM_videoFilters.cpp
+++ b/avidemux/common/ADM_videoFilter2/src/ADM_videoFilters.cpp
@@ -43,6 +43,15 @@ ADM_coreVideoFilter     *ADM_vf_getInstance(int index)
     return ADM_VideoFilters[index].instance;
 }
 /**
+    \fn ADM_vf_getEnabled
+    \brief Return filter enabled status  of rank index
+*/
+bool ADM_vf_getEnabled(int index)
+{
+    ADM_assert(index<ADM_vf_getSize());
+    return ADM_VideoFilters[index].enabled;
+}
+/**
     \fn ADM_vf_getTag
     \brief Return tag of index active filter
 */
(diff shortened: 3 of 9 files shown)


eumagga0x2a:
Thank you, I wonder whether it is too confusing that the last filter in the chain can be disabled, but the output of the chain will always include this filter. This might be related to the concept of the preview in the filter manager which shows the output of the chain up to and including the currently selected filter. How should this be handled when the currently selected filter is disabled?

1. Disable a filter
2. Preview the output of the remaining items of the chain
3. Re-enable the filter

As the design goal is to preserve filter configuration, is this cycle guaranteed to be non-destructive? Can the step 2 alter the configuration of remaining active filters?

Last but not least, how should we call the active filter list once it can include inactive / disabled filters?

> What i dont know is, the scriptEngines.

Do you mean that current interfaces don't know anything about filters which should be added, but made disabled?

szlldm:
The last disabled filter indeed included, it is clearly a bug, i missed that.
The filter manager preview shouled display the filtered result even when the currently selected filter is disabled. Its a feature not a bug. It can be useful if someone want to compare two filter output, ith the same preceeding chain.

> Can the step 2 alter the configuration of remaining active filters?

Did you mean 3. Re-enable the filter? In some cases it is inevitable, as would be deleting and later added again.

> Last but not least, how should we call the active filter list once it can include inactive / disabled filters?

Filter chain?

> Do you mean that current interfaces don't know anything about filters which should be added, but made disabled?

Yes.

This feature clearly needs more refinement :)

eumagga0x2a:
I'm sorry for being slow these days as my spare time I can afford to spend on computer stuff is currently very limited.

> The filter manager preview shouled display the filtered result even when the currently selected filter is disabled. Its a feature not a bug.

Sure, but in case of the selected disabled filter being the last in the chain, this introduces a discrepancy between the preview in the filter manager (which includes the disabled filter) and the preview in the main window (which excludes it). Both behave correctly in their own right, but the result feels inherently buggy for me.

> Did you mean 3. Re-enable the filter?

No, I meant step 2, but I was partially wrong.

Many filters don't validate configuration during contruction phase, preview is safe for them. But quite a few filters validate configuration when the configuration dialog is opened. Replace Step 2 with "Configure a filter farther down in the filter chain" and this step may become destructive.

1. Load a 1280x720 video with a logo in the top left corner
2. Rotate it to 270° (this moves the logo to the bottom left corner)
3. Add mpdelogo2, configure it to remove the logo in the bottom left corner
4. Preview the final result
5. Disable rotate
6. Open mpdelogo2 configuration dialog, accept it.
7. Re-enable rotate
8. Preview the final result

Actual results: the picture at step 4 and at 8 is not the same.

This is nothing new as moving filters around can have the same effect, but this points to what I perceive to be the real issue which needs a solution (with the premise behind the feature to allow to disable active filters taking at symptoms): it would be great to have a proper undo/redo functionality for the filter chain similar to segment layout in the editor. This would naturally provide the feature to disable a filter by simply deleting it as this step will be immediately reversible.

> > Last but not least, how should we call the active filter list once it can include inactive / disabled filters?
> 
> Filter chain?

Not quite, as the filter chain needs to be created without disabled filters...

szlldm:
I still dont see the problem here. In your example i can change the rotation amount, or simply delete the filter, and it will bring the same outcome.
The undo/redo functionality would be great, but it is beyond my skill level. Besides that, it is not the same as the enable/disable functionality.

An example workflow where i would like this disable option (and undo/redo would'nt work):
There are frequent camera switching in a video, one camera angle is dark, the other not so much. Also i want cropping, then fit to (a common) size. I save the segments with a mezzanine encoder, then at the end i append them then encode with the final codec.
Now after every few seconds i need to readjust the crop (large displacement), delete the color/intensity adjuster filters, and add another filter. With a disable option i can use two exclusive crop filter, where only minor adjustment would needed, and i could keep the color filters for each scenario.

> Sure, but in case of the selected disabled filter being the last in the chain, this introduces a discrepancy between the preview in the filter manager (which includes the disabled filter) and the preview in the main window (which excludes it). Both behave correctly in their own right, but the result feels inherently buggy for me.

The preview in the filter manager does not show the complete chain output, but the selected filter's.

eumagga0x2a:
> In your example i can change the rotation amount, or simply delete the filter, and it will bring the same outcome.

This is sort of expected when filters up the chain are reconfigured, but disabling and re-enabling suggests non-destructive, fully reversible operation, a promise it cannot keep.

> The undo/redo functionality would be great, but it is beyond my skill level.

Your skill level is formidable. In a very rough draft of undo/redo functionality, we would probably need a memento in form of a container with items consisting of filter tag (uint32_t) and CONFcouple array for each filter in the chain. Adding/deleting a filter or altering filter configuration would create a new memento and pushe it onto undo stack. undo() would create another memento and pushe it onto redo stack, then recreate filter chain from the memento popped from the undo stack. redo() would mirror this procedure. We would need two actions connected to undo and redo buttons with appropriate keyboard shortcuts.

> An example workflow where i would like this disable option (and undo/redo would'nt work):
> There are frequent camera switching in a video, one camera angle is dark, the other not so much. Also i want cropping, then fit to (a common) size. I save the segments with a mezzanine encoder, then at the end i append them then encode with the final codec.
> Now after every few seconds i need to readjust the crop (large displacement), delete the color/intensity adjuster filters, and add another filter. With a disable option i can use two exclusive crop filter, where only minor adjustment would needed, and i could keep the color filters for each scenario.

This scenario is indeed much more complex than anything I could imagine. The enable/disable option would serve as a replacement for missing preset functionality then (both are not mutually exclusive).

> The preview in the filter manager does not show the complete chain output, but the selected filter's.

It shows the output of the chain up to and including the selected filter (by default the last one). Actually, I figured this out a few years ago as a big surprise by reading code. I guess that adding the name of the selected filter to the window title would improve discoverability.

However, the new implementation creates IMHO a confusing element of not showing the actual output of the filter chain by default if the last filter is disabled (and it would never show the actual output if all active filters are disabled). Unsure how to solve this. Maybe by making the button always preview the final output, with preview of filters higher up in the chain available only via context menu? Could be perceived as a regression, I fear.

avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_seekablePreview.h
```diff
@@ -34,7 +34,7 @@ class Ui_seekablePreviewWindow : public QDialog
 	Ui_seekablePreviewDialog ui;
         
 public:        
-                Ui_seekablePreviewWindow(QWidget *parent, ADM_coreVideoFilter *videoStream, uint32_t defaultFrame = 0);
+                Ui_seekablePreviewWindow(QWidget *parent, ADM_coreVideoFilter *videoStream, QString const& title=QString(""), uint32_t defaultFrame = 0);
```
eumagga0x2a, line 37: I wonder why not just calling `previewDialog->setWindowTitle(title)` from `filtermainWindow::preview`?

eumagga0x2a:
The plan is to wait until new nightlies are generated, then proceed with this one.

avidemux/qt4/ADM_userInterfaces/ADM_filters/Q_seekablePreview.h
szlldm: indeed

eumagga0x2a:
Thank you, we'll see how this is going to evolve.

PR #258, opened 2021-03-30
[artVignette] add coupled spinboxes

szlldm:
Is this okay qt signalingwise?


eumagga0x2a:
> Is this okay qt signalingwise?

Yes, looks good, thank you.

PR #259, opened 2021-03-31
Add spinboxes

szlldm:
and setTabOrder where needed


eumagga0x2a:
A nice usability improvement, thanks.

I wonder whether we need a slider at all when the range is very small like in artGrid, but it does no harm when it is there.

PR #262, opened 2021-04-07
[crop/qt5] improvements

szlldm:
1) Replace green blanking with green alpha blending.
It makes repositioning easier.
It also provides visible guide lines (bug turned out to be a feature :-) ).

2) Prevent dimension jitter on rubber band move.


eumagga0x2a:
Probably not worth trying to handle the rare case of source with odd dimensions in the rubber control.

Thank you very much.

PR #263, opened 2021-04-10
[ADM_preview] better preview dimensions set method, when ZOOM_AUTO

szlldm:


eumagga0x2a:
Thank you.

I never tried to verify whether assumed performance benefits of zoom levels being equal a pow2 were real. With 8k videos, the old implementation was insufficient anyway, thus this policy change makes sense.

PR #265, opened 2021-04-14
[fadeTo] alleviate erratic preview behaviour

szlldm:
When seeking in the preview, and seek into the filter scope (its a keyframe, but not the acutal first frame), it will stuck with the wrong frame.
This is not an absolute fix, just an alleviation.

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/fadeTo/ADM_vidFadeTo.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/fadeTo/ADM_vidFadeTo.cpp
@@ -229,6 +229,11 @@ bool AVDM_FadeTo::getNextFrame(uint32_t *fn,ADMImage *image)
     }
     if( out_of_scope || !first)
     {
+        if(first)
+        {
+            delete first;
+            first=NULL;
+        }
         image->duplicate(next);
         nextFrame++;
         vidCache->unlockAll();


eumagga0x2a:
Thank you for finding the issue. Indeed, we cannot do much when we have sought to a frame past the start of the range.

In general, if something breaks on seek, the problem usually exists in the goToTime() implementation or in the lack thereof as here. Not taking due to wrong approach, but really thanks a lot for the bug report.

szlldm:
Why?
I try to rephrase the problem: the filter takes the first frame of the specified time window, when the play/saving process enter into the window. Then it will use this picture for blending. The issue only emerges, when the user seek in the preview of the filter chain.
During a linear process (eg. transcoding) it will work fine, even with the existing code.
I dont see another solution. If somehow you manage it to jump to the fist frame of the time window to take the frame, it will break the filter chain. The only valid solution would be decoding from the start of the video up to the first frame of the t.w., grab the frame, then reset everything to the beginning, and start decoding and filtering from the beginning.

eumagga0x2a:
The fundamental issue with missing the highway exit remains, of course, but methinks we should just clear the snapshot on seek.

(Well, your approach of clearing the snapshot when not in range has a benefit of easily keeping the good image when seeking within the filter range)

eumagga0x2a:
The problematic scenario:

1. seek past the start of the range (we get a wrong snapshot)
2. preview the output and stop short of the end of the range
3. seek to the start of the range.

As result we keep the wrong snapshot.

szlldm:
It only keeps the wrong snapshot in the preview (transcoding not affected, as filter chain rebuilt), while the above scenario is unlikely. Reopening a preview (instatiating the next whatever filter, or opening the vfm's preview) will seek to zero, and will clear the snapshot (except if the range start at zero PTS).

PR #266, opened 2021-04-15
[ADM_edCache] shared cache

szlldm:
Reduce memory usage when appending multiple files.
Tested with appending 30 short files (4k resolution). The total duration is less than 2 min.
Before: 12 GB RAM usage
With the commit: 6 GB RAM usage (so there is still something that eats the memory (decoder?) )

Mixed HW and SW decoded appending needs further confirmation.

PS: i managed to get VDPAU to work with my low end AMD card (no VP9). Appending a HW and a SW decoded file, there is no new error when using this shared cache. 

PS+OT: Appending a HW and a SW decoded file, the SW decoded part not working, the editor shows the last HW decoded frame during seeking/playing in the SW decoded part. Same behaviour with 2.7.6 appimage.

### DIFF
--- a/avidemux/common/ADM_editor/include/ADM_edCache.h
+++ b/avidemux/common/ADM_editor/include/ADM_edCache.h
@@ -22,6 +22,7 @@ typedef struct cacheElem
 {
 	ADMImage *image;
     uint64_t pts;       // If set to ADM_NO_PTS -> unused entry
+    int instance;
 }cacheElem;
 /**
     \class EditorCache
@@ -30,9 +31,12 @@ typedef struct cacheElem
 class EditorCache
 {
 	private :
-			uint32_t     readIndex,writeIndex;
-			cacheElem	 *_elem;
-			uint32_t	_nbImage;
+            static int          _instanceCount;
+            int                 myInstance;
+			static uint32_t     readIndex,writeIndex;
+            static uint32_t     _commonW,_commonH;
+			static cacheElem	*_elem;
+			static uint32_t	    _nbImage;
             void        check(void);
 	public:
                         EditorCache(uint32_t size,uint32_t w, uint32_t h);
--- a/avidemux/common/ADM_editor/src/utils/ADM_edCache.cpp
+++ b/avidemux/common/ADM_editor/src/utils/ADM_edCache.cpp
@@ -25,20 +25,52 @@
 #define aprintf printf
 #define aADM_warning(...) ADM_warning
 #endif
+
+int EditorCache::_instanceCount = 0;
+uint32_t EditorCache::readIndex, EditorCache::writeIndex;
+uint32_t EditorCache::_commonW, EditorCache::_commonH;
+cacheElem *EditorCache::_elem;
+uint32_t  EditorCache::_nbImage;
 /**
     \fn EditorCache
     \brief Constructor
 */
-
 EditorCache::EditorCache(uint32_t size,uint32_t w, uint32_t h)
 {
-    _elem=new cacheElem[size];
-    for(uint32_t i=0;i<size;i++)
+    if (size > EDITOR_CACHE_MAX_SIZE)
+        size = EDITOR_CACHE_MAX_SIZE;
+    
+    if (_instanceCount == 0) // first instance
     {
-        _elem[i].image=new ADMImageDefault(w,h);
-        _elem[i].pts=ADM_NO_PTS;
+        _elem=new cacheElem[EDITOR_CACHE_MAX_SIZE];
+        _commonW = w;
+        _commonH = h;
+        
+        for(uint32_t i=0;i<size;i++)
+        {
+            _elem[i].image=new ADMImageDefault(w,h);
+            _elem[i].pts=ADM_NO_PTS;
+        }
+        _nbImage=size;
+    } else {
+        ADM_assert(_commonW == w);
+        ADM_assert(_commonH == h);
+        
+        if (size > _nbImage)
+        {
+            for(uint32_t i=_nbImage;i<size;i++)
+            {
+                _elem[i].image=new ADMImageDefault(w,h);
+                _elem[i].pts=ADM_NO_PTS;
+            }
+            _nbImage=size;          
+        }
     }
-    _nbImage=size;
+    
+    flush();
+    
+    myInstance = _instanceCount;
+    _instanceCount++;
     ADM_info("Video cache created for %u decoded images.\n",_nbImage);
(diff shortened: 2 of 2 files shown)


eumagga0x2a:
Feels very risky. I get "unsorted double linked list corrupted" crash on deleting decoder (e.g. when closing Avidemux) attached to an appended video when hw decoder is used:

1. Load a H.264 video decodable with VDPAU
2. Append another H.264 video decodable with VDPAU
3. Start playback, stop playback
4. Close Avidemux = crash.

> so there is still something that eats the memory (decoder?)

Sure, as libavcodec + libavutil keep decoded pictures needed for reference in refcounted buffers.

When hw accel is used, these buffers are hw surfaces pointed to by refHwImage in refDescriptor member of ADMImage in cacheElem. We manage these hw surfaces ourselves and set get_buffer2 callback in AVCodecContext to the pointer to our allocator and the free callback in av_buffer_create to our function to free a surface. Surfaces are recycled / destroyed when our ref counter has reached zero. With a shared cache, it might be that the second (third, etc.) decoder instance finds its surfaces gone.

eumagga0x2a:
> PS+OT: Appending a HW and a SW decoded file, the SW decoded part not working, the editor shows the last HW decoded frame during seeking/playing in the SW decoded part. Same behaviour with 2.7.6 appimage.

Thank you for reminding me, I noticed this bug quite a while ago too, should be high priority.

szlldm:
Now looks like ok.

eumagga0x2a:
> Appending a HW and a SW decoded file, the SW decoded part not working, the editor shows the last HW decoded frame during seeking/playing in the SW decoded part.

Fixed for libavcodec, will fix uncompress() implementation of all other sw decoders later.

> Now looks like ok.

Indeed, thanks, the crash is gone. I don't quite understand why it works yet.

Avidemux was really not using a per-video cache to its potential, I mean it should be possible at least to retain decoded images before a segment switch when the next segment has a different ref video, so that stepping backward across such segment boundaries could be faster.

If we go with a single common cache, what is the purpose of keeping different cache instances per video? Just to store int myInstance? Wouldn't it be more straightforward to move the cache into ADM_EditorSegment and do without any static vars? It makes a slightly bigger patch as the interface needs to be modified, but IMVHO this looks clearer.

[editor-shared-cache-no-static-vars.patch.txt](https://github.com/mean00/avidemux2/files/6328173/editor-shared-cache-no-static-vars.patch.txt)

szlldm:
> I don't quite understand why it works yet.

I'm not 100% sure, but i think the problem was, hw ADM_images were reused without proper cleaning/releasing.

>  I mean it should be possible at least to retain decoded images before a segment switch when the next segment has a different ref video, so that stepping backward across such segment boundaries could be faster.

IMHO it is diminishing return. What would be really nice, if it could retain between segments in the same ref video.

> If we go with a single common cache, what is the purpose of keeping different cache instances per video? Just to store int myInstance? Wouldn't it be more straightforward to move the cache into ADM_EditorSegment and do without any static vars? It makes a slightly bigger patch as the interface needs to be modified, but IMVHO this looks clearer.

I'm not an expert in object oriented programming (i only started using oop and c++ back in january, to improve the usability of this software), but i think this way follows the encapsulation principle  more strictly. And if it turns out to be problematic, reverting is easier.
But i'm not insisting to my solution. If you think the other way better, lets go with that.

PS: my ulterior motive here is to make way a possible increase in EDITOR_CACHE_MAX_SIZE, and consequently to the increase  of the upper limit of the editor cache size preference :-)

eumagga0x2a:
>  only started using oop and c++ back in january

This is pretty amazing.

> And if it turns out to be problematic, reverting is easier.

I'd try to find a solution which can be switched between non-shared and shared mode using a define at compile time or even a preference at runtime.

> my ulterior motive here is to make way a possible increase in EDITOR_CACHE_MAX_SIZE, and consequently to the increase of the upper limit of the editor cache size preference :-)

I fear I have to (partially) disappoint you as the main obstacle – limited hw resources when a hw decoder is used – is not alleviated by sharing a cache, e.g. with VDPAU, we still have to allocate at least cache_size + nb_ref_frames surfaces per decoder instance, i.e. per video. I don't know yet whether we can safely destroy some surfaces (currently we destroy them later when the decoder instance has been deleted) . Even if it turns out that we can do it for VDPAU, this is definitely impossible with DXVA2 where the entire pool of surfaces needs to be allocated upfront.

szlldm:
> I fear I have to (partially) disappoint you as the main obstacle – limited hw resources when a hw decoder is used – is not alleviated by sharing a cache, e.g. with VDPAU, we still have to allocate at least cache_size + nb_ref_frames surfaces per decoder instance, i.e. per video. I don't know yet whether we can safely destroy some surfaces (currently we destroy them later when the decoder instance has been deleted) . Even if it turns out that we can do it for VDPAU, this is definitely impossible with DXVA2 where the entire pool of surfaces needs to be allocated upfront.

Caching matters mostly when using sw decoders. Would it be possible that cache size preference is used only when hw acceleration disabled? HW decoders are pretty fast, and can go with small cache size, so it could be even a hard define.
The optimal cache size would be to fit an entire GOP, but obviously that not realistic. But the current upper limit is too small.

eumagga0x2a:
> Caching matters mostly when using sw decoders. Would it be possible that cache size preference is used only when hw acceleration disabled?

While technically possible, wouldn't it be an awfully awkward design? Especially considering that enabling hw accelerated decoding doesn't mean that a particular video can be decoded in hardware. This would also motivate users to keep hw decoding disabled, which is a bad thing.

> The optimal cache size would be to fit an entire GOP

For optimal results, two GOPs  :-)

> I think it can be achieved inside the cache class, with a boolean parameter at the contructor.

I thought It would go into `ADM_EditorSegment::addReferenceVideo` with a small change to `ADM_EditorSegment::deleteAll`, without EditorCache needing to know anything about its purpose.

szlldm:
> This would also motivate users to keep hw decoding disabled, which is a bad thing.

It would only reduce the gap between the sw vs. hw decoding UX.

> For optimal results, two GOPs

Indeed. Maybe it wouldnt be too difficult to implement two set of cache, one for the current and one for the previous segment.

eumagga0x2a:
I've pushed a solution which implements an optional shared cache, which can be casually switched on and off at runtime, effective for ref videos loaded or appended later. I'll revert the changes if serious issues are found or if future improvements of cache management require that.

szlldm:
Thank you!

PR #268, opened 2021-05-09
Markers...

szlldm:
Makes marker info available for the filters.
Those filters, that change duration, also adjust the marker infos.
Partial filter overwrite marker infos with start stop times, if the son filter opened for configuration.
_The main goal here:_
Filter dialog seeks to markerA upon opening the filter for configuration. Also this means in case of partial filter, it seeks to the start time of the partial filter (regardless of actual editor marker positions).
With the current infrastructure, the seek is only to a key frame.


eumagga0x2a:
Nice, thanks.

Do you plan to find a remedy for filtermainWindow::preview() always seeking to the start of video? It feels inconsistent now as the markers get properly propagated down the filter chain and previews in filter configuration dialogs seek to the A marker.

szlldm:
Ok, i will look into that.

PR #270, opened 2021-05-20
[ADM_flyDialogRgb] improve performance

szlldm:
Resizing or dragging the rubber band in the Crop filter is sluggish at 1080p and even more above.
Therefore i improved on ADM_flyDialogRgb (ADM_flyDialogRgb used only by Crop and Blacken border):
- YUV->RGB conversion is performed only when necessary (Pts of the current image changed)
- above 720p scaling is ADM_CS_FAST_BILINEAR instead of ADM_CS_BICUBIC


eumagga0x2a:
Awesome, thanks. Could you please just replace magic numbers (-1LL and 720) with defines? The first is ADM_NO_PTS, the latter is up to you how to call.

eumagga0x2a:
Thanks!

PR #271, opened 2021-05-20
[asciiView] change category

szlldm:
IMHO the "Artistic" category is a 100% fit for this filter :)


eumagga0x2a:
> IMHO the "Artistic" category is a 100% fit for this filter :)

Agreed.

PR #272, opened 2021-05-20
[videoFilter] New filter: Flip

szlldm:
Horizontal/vertical flips merged into a single filter with preview. 
Also previous horizontalFlip and verticalFlip filters are excluded from the CMake list.
This PR somewhat declutters the filter list of the "Transform" category.


szlldm:
Maybe it would be better, if horizontalFlip and verticalFlip would be also included, but their category changed to VF_HIDDEN.
This way would prevent breaking user's scripts.
What do you think?

eumagga0x2a:
Thanks!

PR #273, opened 2021-05-21
Resizers update

szlldm:


eumagga0x2a:
Great, thank you.

With maximum size raised, do you have an idea what to do with the slider when the size specified by spinboxes exceeds the limit of 200%? It is a sort of a cognitive dissonance to see the percentage stuck at its upper limit while the true magnification is higher.

It doesn't need to be solved right now and in this PR, but I just feel uncomfortable with this detail.

szlldm:
I dont think there is a good solution, or that it should be:

- increase the upper limit to an arbitrary limit (400%, 800% ?)
- rescale the slider (it will jump back and forth as user type in the resolution)
- ?

IMHO the slider is just a convenient tool. Personally i have never used it :)

eumagga0x2a:
I see two possible ways to restore sanity: a) make the slider logarithmic with base derived from the highest possible magnification `= MAXIMUM_SIZE / max(width,height) ` with the center position assigned the raw value of 0 = 100% (may be a bit complicated to implement, no idea whether it will be comfortable to use) or b) remove the slider alltogether.

PR #275, opened 2021-05-25
HiDPI scaling

szlldm:
HiDPI scaling in filter previews,
and main window preview, in case QT simple renderer is used.

### DIFF
--- a/avidemux/common/ADM_render/GUI_simpleRender.cpp
+++ b/avidemux/common/ADM_render/GUI_simpleRender.cpp
@@ -77,6 +77,7 @@ bool simpleRender::displayImage(ADMImage *pic)
     lock.lock();
     int stride=ADM_IMAGE_ALIGN(displayWidth*4);
     myImage=QImage(videoBuffer,displayWidth,displayHeight,stride,QImage::Format_RGB32).copy(0,0,displayWidth,displayHeight);
+    myImage.setDevicePixelRatio(info.scalingFactor);
     lock.unlock();
     refresh();
     return true;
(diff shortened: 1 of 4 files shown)


szlldm:
can be tested with the QT_SCALE_FACTOR environment variable.

eumagga0x2a:
Thanks, I tried a mingw build with this patch on Windows 10 running on a MBR (the only piece of hardware I have access to with a HiDPI display) and it reproducibly crashes with access violation in swscale-5.dll invoked by `ADMColorScalerFull::convertPlanes` when using the Qt video output (all other outputs are fine). Not sure about the direct cause.

A small wish: I would prefer to preserve the ability to build Avidemux with Qt4, which would be broken as QImage in Qt4 doesn't have the `setDevicePixelRatio` method. It should be probably hidden from compiler by a QT_VERSION check.

szlldm:
Unfortunately i can not replicate any crash, but it might be because race condition. I tried to fix it blindly.

eumagga0x2a:
The crash turned out to be my bug, [should be fixed now](https://github.com/mean00/avidemux2/commit/a85ab0d3a648113b47f197d497935624075dabc4).

avidemux/common/ADM_render/GUI_simpleRender.cpp
```diff
@@ -73,10 +73,13 @@ bool simpleRender::refresh(void)
 */
 bool simpleRender::displayImage(ADMImage *pic)
 {
-    scaler->convertImage(pic,videoBuffer);
     lock.lock();
+    scaler->convertImage(pic,videoBuffer);
```
eumagga0x2a, line 77: I think, all changes introduced to avert suspected race condition can be safely reverted now.

eumagga0x2a:
Thank you, will merge now and deal with whatever gets broken on macOS later.

PR #276, opened 2021-05-27
[Q_gui2] resume playing on slider drag

szlldm:
Keep playing if navigation slider has been dragged.


eumagga0x2a:
Very nice, thanks.

PR #277, opened 2021-05-28
[PulseAudioSimple] add volume control

szlldm:
It's a dirty hack, as PulseAudioSimple API does not provides volume control.

OT: Alsa mixer volume control not working on my system, it has only multiple "IEC958" instances, neither "PCM" nor "Master".

### DIFF
--- a/avidemux_plugins/ADM_audioDevices/PulseAudioSimple/ADM_devicePulseSimple.cpp
+++ b/avidemux_plugins/ADM_audioDevices/PulseAudioSimple/ADM_devicePulseSimple.cpp
@@ -38,6 +38,7 @@ pulseSimpleAudioDevice::pulseSimpleAudioDevice()
 {
     instance=NULL;
     latency=0;
+    volumeHack = 32768;
 }
 /**
     \fn pulseSimpleAudioDevice
@@ -147,6 +148,17 @@ pa_channel_map map,*pmap=NULL;
 
 }
 
+
+uint8_t pulseSimpleAudioDevice::setVolume(int volume)
+{
+    if (volume < 0)
+        volume = 0;
+    if (volume > 100)
+        volume = 100;
+    volumeHack = (32768 * volume * volume) / 10000;
+    return 0;
+}
+
 /**
     \fn sendData
     \brief Playback samples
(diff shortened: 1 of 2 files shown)


eumagga0x2a:
> OT: Alsa mixer volume control not working on my system, it has only multiple "IEC958" instances, neither "PCM" nor "Master".

Did you select the right soundcard with the F6 key? Did you try a live system with more recent kernel?

I just realized that when I select an Alsa* output in Avidemux, Avidemux controls the master volume for all applications. I am not sure whether it is a desired behavior or a fixworthy bug. Probably the latter or an artifact of PulseAudio emulation, the volume control for the CoreAudio output on macOS affects the volume level for Avidemux only, similar to this soft volume solution for PAS (which works great).

I feel torn about this change, I personally never missed a volume control in Avidemux with PAS on Linux, rather enjoyed the space freed. Application-specific volume levels can be set in gnome-control-center, they persist at least for a GNOME session.

szlldm:
I checked, so i have 2 audio device, the first is the HDMI output of the VGA, the second is the motherboard audio. The HDMI has those IEC958 (S/PDIF) outputs. Nonethless audio works, just the volume control wont.
I highly prefer setting application-specific volume level control from inside the very same application.

>  I personally never missed a volume control in Avidemux with PAS on Linux, rather enjoyed the space freed

You can hide it, but it is the least wide element of the main window.

eumagga0x2a:
> I checked, so i have 2 audio device, the first is the HDMI output of the VGA, the second is the motherboard audio. The HDMI has those IEC958 (S/PDIF) outputs.

Almost the same here. I have 3 audio devices, the default is the PulseAudio with just the Master and Capture. The second is the motherboard audio which offers a plethora of working controls and the third one is HDA NVidia which offers just a bunch of S/PDIF outputs. My boxes are connected to the line-out of the motherboard, tiny speakers in the display (and thus the HDMI audio output) work, but are useless.

> You can hide it, but it is the least wide element of the main window.

Sure, but this is not fun as long as the widget visibility is not preserved across Avidemux sessions.

szlldm:
> Sure, but this is not fun as long as the widget visibility is not preserved across Avidemux sessions.

IMHO volume control is an essential functionality, and it takes less space than the less useful (but not useless) VU meter.

eumagga0x2a:
I've copypasted from mpv a primitive PulseAudio plugin with a native volume control. If it proves satisfactory, I would disable the build of the old one which uses the simple API.

szlldm:
> I've copypasted from mpv a primitive PulseAudio plugin with a native volume control.

Does it worth the extra work? Its fine by me, just asking. :-)

eumagga0x2a:
> Does it worth the extra work?

I think yes, definitely, it is. Purely from a self-educational POV.

The next step could be disabling or removing obsolete audio output plugins like OSS and ESD.

szlldm:
I looks like, its works. Thank you!

PR #278, opened 2021-05-28
[T_vumeter] extend with bar width and overdrive level parameters

szlldm:
imho the vu meter bars were too thin

### DIFF
--- a/avidemux/qt4/ADM_userInterfaces/ADM_gui/T_vumeter.cpp
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_gui/T_vumeter.cpp
@@ -37,6 +37,9 @@ static ADM_Qvumeter *vuWidget = NULL;
 #define RED    0xFFFF0000
 #define BLACK  0xFF000000
 
+#define BAR_WIDTH  7
+#define OVERDRIVE_LEVEL  80
+
 ADM_Qvumeter::ADM_Qvumeter(QWidget *z, int width, int height) : QWidget(z)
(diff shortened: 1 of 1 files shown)


butterw:
The vu meter bars were too thin, but it is possibly not the main issue.

Cutting the video to audio (ex: I need to edit out some video frames because of unwanted sound) is a struggle with Avidemux currently. The VU doesn't display much useful information when the video is stopped (audio players seems to do a lot better than Avidemux here). Something like a rolling average could be used, so a past loud still shows up in the vu on stopped video but with lesser amplitude.

eumagga0x2a:
> The vu meter bars were too thin

I played with making them thicker quite a long time ago and discarded such attempts for visual reasons (probably personal taste issues). The bars must be twice as thick pixel-wise in HiDPI conditions. If they are not, this needs to be fixed.

> but it is possibly not the main issue.

Indeed. The VU meter is a terrible resource hog. On macOS, it consumes more CPU power than decoding 720p@25fps H.264 in software. This was the reason I extended the effect of refresh cap setting to the VU meter, which rendered this widget even less useful.

The next issue is the VU meter being ahead of the audible sound due to internal buffering.

Indicated values are average (actually, mean square) over the last 5 ms, a peak or clipping detector might be a nice addition once the fundamental issues are solved.

szlldm:
This PR is just an aesthetical improvment. With the widened bars it looks more similar to an actual LED bar VU meter.

PR #279, opened 2021-05-31
[Alsa] replace volume control

szlldm:
I have replaced the volume control with hacked soft control (also fixed the indentations).
The original setVolume was a mess:

- always used (at _snd_mixer_attach_) "hw:0", where ADEVICE should have been used
- however a comment ("_MEANX: Cannot use the real name, does not work with dmix_") indicate it may wont work with "dmix" (i can not confirm it, alsa dmix audio output does not work on my system)
- using ADEVICE, the volume control works, but it will change the global (==system-wide) volume, which is unacceptable.

I have also decreased the buffer size (1s -> 100ms), to reduce the lag between volume change action and effect.

### DIFF
--- a/avidemux_plugins/ADM_audioDevices/Alsa/ADM_deviceALSA.cpp
+++ b/avidemux_plugins/ADM_audioDevices/Alsa/ADM_deviceALSA.cpp
@@ -58,11 +58,19 @@ ADM_DECLARE_AUDIODEVICE(AlsaDefault,alsaAudioDevice,1,0,0,"Alsa Audio Device (de
 /* Handle for the PCM device */
 snd_pcm_t *pcm_handle;
 
-    alsaAudioDevice::alsaAudioDevice( void )
-    {
-        _init=0;
-        volumeHack=0;
-    }
+alsaAudioDevice::alsaAudioDevice( void )
+{
+    _init=0;
+    volumeHack=0;
+    softVolumeBuffer = NULL;
+}
+
+alsaAudioDevice::~alsaAudioDevice( void )
+{
+    if (softVolumeBuffer)
+        delete [] softVolumeBuffer;
+    softVolumeBuffer = NULL;
+}
 /**
     \fn localInit
     \brief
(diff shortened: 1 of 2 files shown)


eumagga0x2a:
Do we have a race condition between `sendData()` and `audioDeviceThreaded::getVolumeStats` if data in the buffer both functions operate on gets altered in-place?

szlldm:
Soft volume application is protected by the mutex, so strictly speaking there is no race condition.
But there is a probability indeed, that getVolumeStats will determine an inaccurate metric.
This could be solved by adding a buffer to the Alsa device.

eumagga0x2a:
I tried to look into ALSA stuff. There are almost no descriptions or examples in the header files, similar to Avidemux. Very hard to use.

It turns out dmix expects SND_PCM_FORMAT_S32 only. I wonder whether this is specific to particular hardware or to ALSA version, but this explains why the dmix output doesn't work.

The original code in `alsaAudioDevice::sendData` doesn't seem to handle a theoretically valid scenario that `snd_pcm_writei` doesn't consume all the data passed to it in the buffer. I have no idea whether it is safe to cut this corner.

In most cases the volume is 100% or 0% (muted), does it bring any performance benefits to check for that and skip softvolume code?

szlldm:
> The original code in alsaAudioDevice::sendData doesn't seem to handle a theoretically valid scenario that snd_pcm_writei doesn't consume all the data passed to it in the buffer. I have no idea whether it is safe to cut this corner.

I did not see complaints about that, so i assume it is safe.
It is set to blocking mode, and according to the manual:
_If the blocking behaviour is selected and it is running, then routine waits until all requested frames are played or put to the playback ring buffer. The returned number of frames can be less only if a signal or underrun occurred._

> In most cases the volume is 100% or 0% (muted), does it bring any performance benefits to check for that and skip softvolume code?

I will look into that.

eumagga0x2a:
Thank you for explaining why partial consumption of input data doesn't need to be handled.

Regarding the PR, there is an extraneous `mutex.unlock()` call in `alsaAudioDevice::sendData()` causing a flood of threading error messages. I'll merge once the mutex thing is fixed, thank you.

Only alsaDefault output (i.e. PulseAudio) works on my system, but this is not related to your changes.

eumagga0x2a:
Thanks!

PR #280, opened 2021-06-01
[ADM_vidPartial] get marker info from the filter chain

szlldm:
take start/stop times from the filter chain


eumagga0x2a:
Yes, indeed. Thanks!

PR #281, opened 2021-06-01
[GUI] save widgets visibility preferences

szlldm:


eumagga0x2a:
Thank you, looks great. Sure, this is the most straightforward way to get the job done and we have quite a few precedents in filter configuration, but did you evaluate using [QMainWindow::saveState](https://doc.qt.io/qt-5/qmainwindow.html#saveState) for everything the main window related? This would automatically cover not only the visibility of dock widgets, but additionally the position of the toolbar and the position and size of the window itself.

eumagga0x2a:
The example following the description of `QMainWindow::saveState` uses QSettings to store the returned QByteArray, this would bypass our own configuration infrastructure completely, if I am not mistaken. Not that the GTK GUI is about to be resurrected any time soon, but IMHO not tying the configuration to a particular GUI implementation and layout is a virtue.

Of course, making window size and position persist means changing the UI paradigm of Avidemux, but why not?

szlldm:
I have tried the suggested Qt stuff, and i did not liked it:
- restoring the windows size is bad, if you had large window (but not maximized), restoring a large window without video looks goofy.
- widget visibility is restored as expected, but in the View Menu / Toolbars the checkboxes are always checked after start.
- restoring a state is sluggish, first you can see the normal sized window, for almost half a second, then it pop to the saved size.

eumagga0x2a:
>   restoring the windows size is bad, if you had large window (but not maximized), restoring a large window without video looks goofy.

This is very valid. As long as window size follows zoom level instead of adding scroll bars, we get all sorts of inconsistent and confusing behaviour.

>   widget visibility is restored as expected, but in the View Menu / Toolbars the checkboxes are always checked after start.

Could you please share the code? We might do something Qt doesn't expect. It can be also a Qt bug, in which case it would be interesting to see whether it persists with Qt6.

>   restoring a state is sluggish, first you can see the normal sized window, for almost half a second, then it pop to the saved size.

This might be related to us showing the main window in `initGUI()` prior to starting event loop in `UI_RunApp()` and doing hw accel probing and plugin loading in-between, which takes quite a lot of time.

szlldm:
> Could you please share the code?

Well, i have scrapped those changes, but it wasnt very complicated:
For loading:
`QSettings* settings = new QSettings(QDir::currentPath() + "/my_config_file.ini", QSettings::IniFormat);
restoreGeometry(settings->value("geometry").toByteArray());
restoreState(settings->value("windowState").toByteArray());`

For saving:
`QSettings* settings = new QSettings(QDir::currentPath() + "/my_config_file.ini", QSettings::IniFormat);
settings->setValue("geometry", saveGeometry());
settings->setValue("windowState", saveState());
settings->sync();`

I have used them at the same location in UI_RunApp and UI_closeGui, as in my commit.

> This might be related to us showing the main window in initGUI() prior to starting event loop in UI_RunApp() and doing hw accel probing and plugin loading in-between, which takes quite a lot of time.

Loading can not happen earlier, because some other part would override the loaded state (volume control visibility for example).

So to get it to work properly, IMHO a lot of changes should have been made, to achive miniscule gain over the proposed solution....

eumagga0x2a:
> Loading can not happen earlier, because some other part would override the loaded state (volume control visibility for example).

We would need to stop forcibly hiding the volume widget when the selected audio device provides no volume control then (disabling is okay).

If we don't touch window geometry, we can restore the window state already in `initGUI`, e.g.:

```
diff --git a/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp b/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
index dbaa55361..7bc59b6cb 100644
--- a/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
@@ -20,6 +20,7 @@
 #include <QGraphicsView>
 #include <QtCore/QDir>
 #include <QMessageBox>
+#include <QSettings>
 #if QT_VERSION < QT_VERSION_CHECK(5,11,0)
 #   include <QDesktopWidget>
 #else
@@ -1875,18 +1876,26 @@ void MainWindow::nextIntraFrame(void)
 */
 void MainWindow::volumeWidgetOperational(void)
 {
-    // Hide and disable the volume widget if the audio device doesn't support setting volume
-    if(!AVDM_hasVolumeControl())
-    {
-        ui.volumeWidget->setEnabled(false);
-        ui.volumeWidget->hide();
-        ui.actionViewVolume->setChecked(false);
-    }else
-    {
-        ui.volumeWidget->setEnabled(true);
-        ui.volumeWidget->show();
-        ui.actionViewVolume->setChecked(true);
-    }
+    // Disable the volume widget if the audio device doesn't support setting volume
+    ui.volumeWidget->setEnabled(AVDM_hasVolumeControl());
+}
+
+/**
+    \fn syncToolbarsMenu
+    \brief Make sure only visible widgets have check marks
+           in the Toolbars submenu of the View menu.
+*/
+void MainWindow::syncToolbarsMenu(void)
+{
+#define EXPAND(x) ui.x ## Widget
+#define CHECKMARK(x,y) ui.menuToolbars->actions().at(x)->setChecked(EXPAND(y)->isVisible());
+    CHECKMARK(0,audioMetre)
+    CHECKMARK(1,codec)
+    CHECKMARK(2,navigation)
+    CHECKMARK(3,selection)
+    CHECKMARK(4,volume)
+#undef CHECKMARK
+#undef EXPAND
 }
 
 MainWindow::~MainWindow()
@@ -1939,6 +1948,10 @@ int UI_Init(int nargc, char **nargv)
     global_argc=nargc;
     global_argv=nargv;
     ADM_renderLibInit(&UI_Hooks);
+
+    QCoreApplication::setOrganizationName("Avidemux");
+    QCoreApplication::setOrganizationDomain("avidemux.org");
+    QCoreApplication::setApplicationName("Avidemux");
 #if !defined(__APPLE__) && QT_VERSION >= QT_VERSION_CHECK(5,11,0) && QT_VERSION < QT_VERSION_CHECK(6,0,0)
     // Despite HiDPI scaling being supported from Qt 5.6 on, important aspects
     // like OpenGL support were fixed only in much later versions.
@@ -1977,7 +1990,10 @@ int UI_Init(int nargc, char **nargv)
 uint8_t initGUI(const vector<IScriptEngine*>& scriptEngines)
 {
     MainWindow *mw = new MainWindow(scriptEngines);
+    QSettings qset;
+    mw->restoreState(qset.value("windowState").toByteArray());
     mw->show();
+    mw->syncToolbarsMenu();
 
     QuiMainWindows = (QWidget*)mw;
 
@@ -2025,12 +2041,14 @@ uint8_t initGUI(const vector<IScriptEngine*>& scriptEngines)
  */
 void UI_closeGui(void)
 {
-        if(!uiRunning) return;
-        uiRunning=false;
-        
+    if(!uiRunning) return;
+    uiRunning=false;
+
+    QSettings qset;
+    qset.setValue("windowState", ((QMainWindow *)QuiMainWindows)->saveState());
+
     QuiMainWindows->close();
     qtUnregisterDialog(QuiMainWindows);
-        
 }
 
 void destroyGUI(void)
diff --git a/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.h b/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.h
index e8a005331..57f15f813 100644
--- a/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.h
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.h
@@ -116,6 +116,7 @@ public:
     void setResizeThreshold(int value);
     void setActZoomCalledFlag(bool called);
     void setZoomToFit(void);
+    void syncToolbarsMenu(void);
     static void updateCheckDone(int version, const std::string &date, const std::string &downloadLink);
     static MainWindow *mainWindowSingleton;
 
```
`syncToolbarsMenu()` overrides the hardcoded "checked" property set in gui2.ui, which is the reason for the checkboxes being always checked upon application restart.

eumagga0x2a:
Portable mode doesn't allow to let QSettings decide which format to take and where to store the data, I would need to test on Windows the impact of a non-native format.

In general, QSettings seem to work great when GUI-specific configuration needs to be saved only per class rather than per instance. What do you think about the most suitable policy for visibility of rubber bands in various filters? If you hide it e.g. in the crop filter, do you expect it to be hidden all crop filter instances once you open the configuration dialog?

szlldm:
> What do you think about the most suitable policy for visibility of rubber bands in various filters? If you hide it e.g. in the crop filter, do you expect it to be hidden in all crop filter instances once you open the configuration dialog?

Its a tough question. Maybe its okay to expect hidden in the next filter instance. However, if i launch adm next time, i would expect it to show the rubber band.

szlldm:
I did noticed your new commit, there is a serious issue with it:
once you detach the toolbar (the one just under the menus), you can not (at least i cant) get docked again. Obviously no restart will fix it.

eumagga0x2a:
> once you detach the toolbar (the one just under the menus), you can not (at least i cant) get docked again.

Resize the window to increase its height. The toolbar can be docked again then.

The whole is a WISP (work in slow progress), a function to reset GUI to default without manually deleting QtSettings.ini while Avidemux is not running is coming.

eumagga0x2a:
> Maybe its okay to expect hidden in the next filter instance. However, if i launch adm next time, i would expect it to show the rubber band.

I struggle to understand why rubber band visibility in the crop filter should be different (should be reset) after application restart (even if it would be trivial to implement it). However, there is indeed a problem of the rubber band visibility being merely a usage hint in some filters (crop, blackenBorder) and an essential functionality in others (blur, mplayerDelogo). This means there is probably no "one size fits all" solution.

I've pushed two experimental commits to use QSettings for rubber band visibility in crop and blackenBorder, thus opening the Pandora's Box of inconsistency. I might revert both, haven't made my mind yet.

### OT:

For something completely different: Would you like to join the Avidemux team on Slack for easier communication? I would try to send you the invitation for the email address mentioned in your commit headers then.

szlldm:
> I struggle to understand why rubber band visibility in the crop filter should be different (should be reset) after application restart

In an editing session its o.k. because i will remember if i have turned if off. But next time i open the software it is shocking to me, when i open the crop filter and i dont see the rubber band.

> Would you like to join the Avidemux team on Slack for easier communication?
O.k. However i will have less free time for this purpose in the future as it looks like things are turning back to normal again after this pandemic turmoil.

eumagga0x2a:
> In an editing session its o.k. because i will remember if i have turned if off. But next time i open the software it is shocking to me, when i open the crop filter and i dont see the rubber band.

Done, quick & dirty style, export `ADM_QT_DROP_SETTINGS=1` to get the desired behaviour.

> O.k. However i will have less free time for this purpose in the future as it looks like things are turning back to normal again after this pandemic turmoil.

Not much different here. The invitation should have been sent by now.

PR #283, opened 2021-06-02
[GUI] buttons and shortcuts for jump to cut point

szlldm:
Shortcuts (Shift + Up/Down) or buttons (with new icons) can be used to jumpt to cut points (indicated on the navigation slider)


eumagga0x2a:
Thank you.

PR #285, opened 2021-06-09
[resampleFps] add motion compensated interpolation option

szlldm:
its experimental / far from perfect, it does not handle large motion well...
it is significantly faster than ffmpeg's minterpolate, but quality lacks behind.
however, if there are no large motions (eg. 20fps->60fps, 30fps->60fps) it is better most of the time than the "Blend" interpolation.
so dont try 1..2..5 fps->60fps conversion :)
there is no scene change detection. visible artifacts can occur.
overall, its not bad for a two afternoon job :)

i dont know cmake at all, so it was a little struggle to make CMakeLists.txt work, i hope its okay in the current state.

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/resampleFps/ADM_vidResampleFPS.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/resampleFps/ADM_vidResampleFPS.cpp
@@ -24,6 +24,7 @@
 
 #include "confResampleFps.h"
 #include "confResampleFps_desc.cpp"
+#include "motin.h"
 
 #if 1
     #define aprintf(...) {}
@@ -63,6 +64,7 @@ protected:
         ADMImage            *frames[2];
         bool                refill(void);   // Fetch next frame
         bool                prefillDone;        // If true we already have 2 frames fetched
+        motin *             motinp;
 public:
                             resampleFps(ADM_coreVideoFilter *previous,CONFcouple *conf);
                             ~resampleFps();
@@ -105,7 +107,23 @@ bool resampleFps::updateIncrement(void)
 const char *resampleFps::getConfiguration( void )
 {
 static char buf[256];
- snprintf(buf,255," Resample %sto %2.2f fps",(configuration.blend ? "with blending ":""),(double)configuration.newFpsNum/configuration.newFpsDen);
+ const char * intpn = NULL;
+ switch (configuration.interpolation)
+ {
+     case 0:
+         intpn = "none";
+         break;
+     case 1:
+         intpn = "blend";
+         break;
+     case 2:
+         intpn = "motion compensation";
+         break;
+     default:
+         intpn = "INVALID";
+         break;
+ }
+ snprintf(buf,255," Resample to %2.2f fps. Interpolation: %s", (double)configuration.newFpsNum/configuration.newFpsDen, intpn);
  return buf;  
 }
 /**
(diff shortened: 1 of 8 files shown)


szlldm:
Tested on the same 720p@29.97fps sample (upsampled to 60fps), this is 9 times faster than ffmpeg's minterp. Quality difference is not significant (but sometimes noticable if i watch especially for differences).

eumagga0x2a:
I apologize for the delays.

I use

```
INCLUDE(vf_plugin)

SET(ADM_vf_resampleFps_SRCS ADM_vidResampleFPS.cpp motin.cpp)

IF(DO_COMMON)
    include(admCheckThreads)
ENDIF(DO_COMMON)

ADD_VIDEO_FILTER(ADM_vf_resampleFps ${ADM_vf_resampleFps_SRCS})
INIT_VIDEO_FILTER(ADM_vf_resampleFps)
INSTALL_VIDEO_FILTER(ADM_vf_resampleFps)

IF(DO_COMMON)
    ADM_LINK_THREAD(ADM_vf_resampleFps)
ENDIF(DO_COMMON)
```
as CMakeLists.txt to build the enhanced resampleFps filter. I haven't tried cross-compilation and building on macOS yet (and I cannot test MSVC, obviously).

The preference FEATURES_THREADING_LAVC is not meant to control anything beyond the scope hinted by its name, i.e. it should not be used to control multi-threading in components other than libavcodec-based video decoders. It is also currently always set to 1 (single thread) when hw accelerated decoding is used.

> overall, its not bad for a two afternoon job :)

It is outright amazing. However, the performance with a 654x480 sample and multi-threading hardcoded to 2 threads (my PC has a dual-core CPU) is 10 fps and 8 fps with multi-threading disabled. My feeling is that when it is that slow, it doesn't matter that it is faster than minterpolate. Personally, I would prefer going for maximum quality no matter how slow, else it feels like finding a middle ground between exiting a highway and continuing straight ahead.

szlldm:
Thanks for the cmake file!

> However, the performance with a 654x480 sample and multi-threading hardcoded to 2 threads (my PC has a dual-core CPU) is 10 fps and 8 fps with multi-threading disabled.

I had around 35 fps at the mentioned 720p (29.97->60 fps) conversion. With the new commit, i can get 50 fps.

> My feeling is that when it is that slow, it doesn't matter that it is faster than minterpolate. 
IMHO 3-4 fps vs 50 fps is a BIG difference.

> Personally, I would prefer going for maximum quality no matter how slow,

My first attempt was, to port/use minterpolate, but it would be very hard. If someone will port that, it could be a 4th option, and the user could decide between a fast and a slow (but better quality) motion interpolation.

>  else it feels like finding a middle ground between exiting a highway and continuing straight ahead.
Continuing the "car metaphore", quality-wise the Blend option is a wheelchair, my "Motin" is a scooter, and minterpolate is a regular car. (Fps-wise its in reverse :) )

szlldm:
I had improved on quality. Now there is no big difference on the sample i used.
I have also tried larger upsampling, 29.97->180 fps is 20x faster than minterpolate.
So why 180fps? Chaining resampleFps and changeFps (29.97 -> 180 -> 30 or 60)  you can do slow motion effect.
I will try to add a scene change detector to improve further.

szlldm:
With a simple trick a could improve the quality a large extent. Now its basically on par with minterpolate in quality (compared on a very limited sample set, due to minterpolate's extreme slowness).

szlldm:
DONE!
I have fixed the motion estimation errors at the edges. Now i cant perceive any differences between minterpolate and motin (tested on a section of the famous Costa Rica 4k video, downsized to 480p before get slowmotioned).

eumagga0x2a:
Thank you very much. I pulled it, we'll see what breaks. Please keep maintaining the new feature, I cannot do it for you :-)

The quality is indeed better, it is worth a drop of speed to 3 fps with my sample, IMHO.

Of course, all the motion estimation cannot perform miracles, i.e. handle the image of moving bow of a string player rapidly changing strings back and forth. When resampled from almost 1001/30000 to strictly 1001/30000 with ME, you see two bow tips. In the original video, there is no temporal bleding, all clean :-)

> Continuing the "car metaphore"

There was actually no car metaphore involved, purely the problem of policy. An example of a situation where an agreement based on a trade off (without broadening the scope of negotiotions) is much worse than no agreement at all.

PR #286, opened 2021-06-13
[motin] speedup by precalculating distance penalties

szlldm:
I wouldnt tought that std::pow is that slow.
With precalculating i get more than 30% fps increase.


eumagga0x2a:
I get more than twice as high fps values with this patch (7 fps instead of 3 with my [favourite sample](https://www.youtube.com/watch?v=bGQbPi_lbvI) which was a flv file back then in early days of YouTube, with slightly irregular timing; look at the bow tip between seconds 8 and 9), a tremendous improvement!

Thank you very much.

PR #288, opened 2021-06-16
Increase max fps to 1000

szlldm:
+ related bugfix in changeFps


eumagga0x2a:
Thank you for the changeFps fix.

Regarding 1000 as fps limit, I wonder, what is the use case? Is there any hardware around capable of playing high-fps content beyond, say, 240 fps?

szlldm:
> Regarding 1000 as fps limit, I wonder, what is the use case? Is there any hardware around capable of playing high-fps content beyond, say, 240 fps?

There are already displays with 360 Hz refresh rate, so you can think of this limit as future proofing.
Aside from that, if you want to make a slow motion video with motion interpolation, now you can do something like Resample fps (60 -> 960) + Change fps (960 -> 60).

eumagga0x2a:
Oh, I see. Thanks.

PR #289, opened 2021-06-24
[ADM_flyDialog] fix fitCanvasIntoView

szlldm:
The bug only occured with HiDPI scaling & lower (<720p) resolutions videos, and only in filters using ADM_flyDialogRgb.
Symptom: filter's preview is garbage (but obviously looks like scaler - buffer dimension mismatch), which is gone once i resize the filter windows just a small amount.

### DIFF
--- a/avidemux/qt4/ADM_UIs/src/DIA_flyDialog.cpp
+++ b/avidemux/qt4/ADM_UIs/src/DIA_flyDialog.cpp
@@ -698,27 +698,24 @@ void ADM_flyDialog::fitCanvasIntoView(uint32_t width, uint32_t height)
     double ar = (double)_w / _h;
     double viewAr = (double)width / height;
 
-    if(viewAr != ar)
+    uint32_t tmpZoomW = 0;
+    uint32_t tmpZoomH = 0;
+    if(viewAr > ar)
     {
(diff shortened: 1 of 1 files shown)


eumagga0x2a:
Does `devicePixelRatioF()` return different values during configuration dialog construction and when it gets shown (i.e. when the first resize even is fired, we call `fitCanvasIntoView()` and sometimes trip over _rgbByteBufferDisplay size mismatch)?

I wonder whether we could avoid always reallocating the buffer even if the size of canvas doesn't change.

Apart from that, a good catch, thanks!

szlldm:
> Does devicePixelRatioF() return different values

I hope not. But currently i can not test it. 
IMHO the deleted _if(viewAr != ar)_ condition is flawed. (Maybe width==_w  and height==_h would make more sense?)

eumagga0x2a:
> IMHO the deleted _if(viewAr != ar)_ condition is flawed.

This. Absolutely. For some reason, this condition was true only once, at the very first resize event, creating the bug.

I think, avoiding wasteful reallocation of memory can be solved separately.

I have checked `devicePixelRatioF()` in a MinGW build on Windows 10 with a few debug messages added, it always returns the expected, correct value, so that idea is moot.

Thank you.

PR #290, opened 2021-07-01
[videoFilter] New filter: Quadrilateral transformation

szlldm:


eumagga0x2a:
Pulling unchecked, it will be most likely fine.

Thank you.

PR #292, opened 2021-07-04
[crop/qt5] fix autocrop

szlldm:
-works on darker scenes then before
-allow odd values (opposite pairs must have same parity)
-it can achieve perfect crops (no black borders left)
-changed ButtonRole of the Auto Crop button to ResetRole, so it is now groupped with the Reset button

### DIFF
--- a/avidemux_plugins/ADM_videoFilters6/crop/qt5/DIA_flyCrop.cpp
+++ b/avidemux_plugins/ADM_videoFilters6/crop/qt5/DIA_flyCrop.cpp
@@ -646,7 +646,7 @@ Ui_cropWindow::Ui_cropWindow(QWidget* parent, crop *param,ADM_coreVideoFilter *i
     connect(pushButtonReset,SIGNAL(clicked(bool)),this,SLOT(reset(bool)));
 
     const QString autorun = QT_TRANSLATE_NOOP("crop","Auto Crop");
-    pushButtonAutoCrop = ui.buttonBox->addButton(autorun,QDialogButtonBox::ResetRole);    // group together with the Reset button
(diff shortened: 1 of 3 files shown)


avidemux_plugins/ADM_videoFilters6/crop/qt5/DIA_flyCrop.cpp
```diff
@@ -625,7 +646,7 @@ Ui_cropWindow::Ui_cropWindow(QWidget* parent, crop *param,ADM_coreVideoFilter *i
     connect(pushButtonReset,SIGNAL(clicked(bool)),this,SLOT(reset(bool)));
 
     const QString autorun = QT_TRANSLATE_NOOP("crop","Auto Crop");
-    pushButtonAutoCrop = ui.buttonBox->addButton(autorun,QDialogButtonBox::ActionRole);
+    pushButtonAutoCrop = ui.buttonBox->addButton(autorun,QDialogButtonBox::ResetRole);    // group together with the Reset button
```
eumagga0x2a, line 649: It would be great not to compromise code semantics to implement minor usability or purely cosmetical tweaks, especially when the perceived issue results from Qt following system guidelines (e.g. QDialogButtonBox::ActionRole is already placed on the left side on macOS). IMHO, it means actually introducing bugs to counteract something which is sometimes unfortunate.

Please revert this one change.

szlldm:
reverted, but not happy with it:
adm_crop

eumagga0x2a:
Thank you very much. IMHO choosing definitions based on their (variable) implementations is a dangerous path. I do acknowledge that the placement like on macOS looks more balanced.

PR #294, opened 2021-07-08
[videoFilter] New filter: Image stabilizer

szlldm:
Reduce camera shakiness.
I hope the cmake file is OK.


eumagga0x2a:
I cannot count how many times this feature was requested, thank you very much.

The only thing which I noticed (apart from encapsulation issues which are probably trivial to fix, so I can do it later myself) and which whould be great to sort out right away is preview changed after "peek original" button pressed and released.

szlldm:
I have removed the "peek original" button, because there is no simple solution for the mentioned problem. IMHO it is not a critical miss.

szlldm:
Probably i have to take the harder route, as the same effect occurs when changing parameters :'(

szlldm:
Done. The "peek original" button will not be returned, as internal state wouldnt be updated if the button would be pressed during a play.

eumagga0x2a:
Thanks!

PR #295, opened 2021-07-13
[Q_gui2] filter navigation actions to prevent congestions

szlldm:
When i continuously press left/right key for navigation (for fine cutting this is inevitable), and the decoder cant keep up, there is a congestion in the action queue. Especially using left key, i can experience more than 10 seconds unresponsive time, when it just keeps going back after i have released the key, and goes way over the desired frame.


eumagga0x2a:
Works great, thanks :-)

PR #296, opened 2021-07-13
[gui_play] improve UI responsiveness during playback

szlldm:
I remember there was issues reported about unresponsive UI during playback (mostly on windoze).
I have tested it by limiting adm to single core and playing demanding content, i saw some delayed (max 1 sec) response to play stop action.
This patch should fix it, but i can not test it on other platform.

### DIFF
--- a/avidemux/common/gui_play.cpp
+++ b/avidemux/common/gui_play.cpp
@@ -255,6 +255,8 @@ bool GUIPlayback::run(void)
 
     uint32_t movieTime;
     uint32_t systemTime;
+    uint32_t lastUIpurgeTime = 0;
+    int32_t delta;
     uint32_t fn;
     bool gotAudio=true;
     bool first=true;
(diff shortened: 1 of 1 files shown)


eumagga0x2a:
> I remember there was issues reported about unresponsive UI during playback (mostly on windoze).

https://avidemux.org/smif/index.php?topic=19231.0

I compared a build containing this patch to a build without it on Linux (NVIDIA), Windows 10 and Windows 7 using a 60 fps 720p sample and didn't see any difference, both were sufficiently responsive. Haven't tested on Linux/Intel yet.

At least, this patch doesn't do any harm. I'll merge it and try to get some feedback once nightlies get refreshed.

szlldm:
I couldnt myself replicate a larger unresponsiveness.
But looking at the code, i could imagine a scenario, where the decoder barely keeps up with the desired rate, then
`if(systemTime>movieTime) // We are late, the current PTS is after current closk`
would be almost never met, because we are just in time, and therefore
`while(delta > 10)`
would also never run, consequently there wouldnt be any _UI_purge()_ call, which is required to QT do its job.

eumagga0x2a:
Indeed, this must have been exactly the reason we ended up sometimes with unresponsive GUI. After a few attempts, I could reproduce the issue on Linux/Intel with an older build, but not with a current one, no matter how hard I tried. Forcing a minimum GUI refresh rate of 30 Hz has fixed the bug.

Thank you very much.

PR #298, opened 2021-07-19
[Q_gui2] show busy cursor during navigation

szlldm:
give feedback during navigation

### DIFF
--- a/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
@@ -448,6 +448,60 @@ void MainWindow::setRefreshCap(void)
 }
 
 /**
+    \fn busyTimerTimeout
+*/
+void MainWindow::busyTimerTimeout(void)
+{
+    if ((busyCntr == 0) && QApplication::overrideCursor())
+        QApplication::restoreOverrideCursor();
+}
+
+/**
+    \fn actionSlot
+*/
+void MainWindow::actionSlot(Action a)
+{
+    if(a==ACT_PlayAvi) // ugly
+    {
+        playing=1-playing;
+        setMenuItemsEnabledState();
+        playing=1-playing;
+    }
+    if(a>ACT_NAVIGATE_BEGIN && a<ACT_NAVIGATE_END)
+    {
+        busyTimer.stop();
+        busyCntr++;
+        if (!QApplication::overrideCursor())
+            QApplication::setOverrideCursor(Qt::WaitCursor);
+    }
+    actionLock++;
+    HandleAction(a);
+    actionLock--;
+    setMenuItemsEnabledState();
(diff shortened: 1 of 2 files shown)


eumagga0x2a:
Great, thank you.

A luxury wish: would it be possible to override cursor only if the wait exceeds some threshold? Keeping e.g. the next frame button pressed produces a lot of flicker from changing the cursor.

szlldm:
> A luxury wish: would it be possible to override cursor only if the wait exceeds some threshold?

That was my first attempt (with QTimer::singleshot), but it does not worked. I suspect the cause is, Qt internal stuff is not updated, while ffmpeg do the magic in the background.

eumagga0x2a:
Indeed, looks like a tough nut to crack. The slot for timeout() signal of a QTimer is not called before actionSlot() is completed.

There is more stuff I don't fully understand yet, e.g. NAVIGATION_ACTION_LOCK_THRESHOLD is almost never exceeded. With an _else_ branch added to the condition `if (actionLock<=NAVIGATION_ACTION_LOCK_THRESHOLD)`, it is almost never reached.

_busyCntr_ gets slowly incremented until it reaches NAVIGATION_ACTION_LOCK_THRESHOLD, then decremented to zero within a millisecond, resulting in cursor flickering briefly every NAVIGATION_ACTION_LOCK_THRESHOLD steps.

szlldm:
> There is more stuff I don't fully understand yet, e.g. NAVIGATION_ACTION_LOCK_THRESHOLD is almost never exceeded. With an else branch added to the condition if (actionLock<=NAVIGATION_ACTION_LOCK_THRESHOLD), it is almost never reached.

That is interesting, because otherwise it stops piling up unserviced key events. 

Maybe the flickering could be eliminated, by adding delay _after_ busyCntr back to zero.

eumagga0x2a:
To drain accumulating signals, we could evaluate the currently unused static admMutex singleThread in `HandleAction()` in gui_main.cpp like

```
    if(action != ACT_PlayAvi &&
       action != ACT_StopAvi &&
       action != ACT_CLOSE &&
       action != ACT_EXIT &&
       singleThread.isLocked())
    {
        ADM_info("mutex locked, returning early.\n");
        return;
    }
```
instead of solving it in Qt.

> Maybe the flickering could be eliminated, by adding delay _after_ busyCntr back to zero.

Could you please elaborate?

szlldm:
> To drain accumulating signals, we could evaluate the currently unused static admMutex singleThread in HandleAction() in gui_main.cpp
instead of solving it in Qt.

That would be a bad idea. On the one hand the mutex is not unused, it is the sole guard that provides the in-order execution of actions; on the other this would discard every action while the sole one is serviced. The NAVIGATION_ACTION_LOCK_THRESHOLD makes possibble to queue a limited number of action (for example if i want to go forward 3 frames, and i press right arrow key three times, using the mutex it could discard 2 keypress).

>>Maybe the flickering could be eliminated, by adding delay after busyCntr back to zero.
> Could you please elaborate?

See the new commit :)

PR #299, opened 2021-07-19
[resize/qt4] display output aspect ratio and nearest common ratio

szlldm:


eumagga0x2a:
Thank you, I find it confusing that the location of labelOutAR is in the group box for pixel aspect ratio while the values displayed there are determined by the controls in the dimensions group box. Being so close to the source PAR combobox creates a wrong impression that the values would reflect the properties of the source. We have space to the right of labelErrorXY aplenty and IMHO would not need the left part of the text displayed in labelOutAR.

szlldm:
Relocated, but not right to labelErrorXY, because it was jerky there (labelErrorXY changes width).

eumagga0x2a:
I think, it should be easy to fix jerkiness by splitting the label in three and forcing fixed width in showEvent() for those with variable content based on font metrics. I can do it myself.

PR #300, opened 2021-07-19
[crop/qt5] display output aspect ratio and nearest common ratio

szlldm:


eumagga0x2a:
Thank you, but I fear that the design may be confusing, especially when selecting a fixed aspect ratio (e.g. 18:9) and watching the label showing (and with small sizes, also sometimes not showing) "2:1". The text also shifts at a high frequency when the rubber band is resized. When the preview dialog size is small, this may also resize the dialog.

Probably splitting the label in three and setting a minimum size for each one based on font metrics would fix the jerkiness. Unsure about the mismatching standard aspect ratio name. Maybe not showing it when a fixed aspect ratio is enabled in the dropdown menu?

eumagga0x2a:
Thank you, I hope users will figure out the meaning of the provided info on their own.

szlldm:
I think those who needs this info, will figure it out.

eumagga0x2a:
It would be better to avoid any ambiguity, but in this case I really didn't have a good solution where to put the labels.

I've added a small patch to be more economical with space.

PR #301, opened 2021-07-22
[Q_gui2] make the Tool Bar hideable too

szlldm:


eumagga0x2a:
Thank you very much, would it be possible to adjust zoom when the available space for the video display changes, please? This affects, of course, the codec widget and the sum of the bottom area widgets as well. No problem with treating this as a separate issue, however.

PS: "Tool Bar" needs to be "Toolbar".

szlldm:
> would it be possible to adjust zoom when the available space for the video display changes, please? 

Just press 0  😁
Otherwise, it would require rework and relocate of the signal handling in .ui into the .cpp file.

eumagga0x2a:
> Otherwise, it would require rework and relocate of the signal handling in .ui into the .cpp file.

Yes, sure. It would be great to automate zoom changes (alternatively, resizing the window when there is enough space) on widgets and toolbar visibility changes for the sake of polished UX. No prob doing that myself at some later point.

PR #305, opened 2021-07-28
[imageStab] performance improvements

szlldm:
- faster fast motion estimation
- apply transform directly on YUV image


eumagga0x2a:
Thank you very much.

PR #307, opened 2021-07-29
[Q_props] display average video bitrate if available

szlldm:
for now only works with mkv/webm and mp4 containers (but these are the most common nowadays)


eumagga0x2a:
Great, thank you. I observed the following bug: when demuxer provides no average bitrate and there are multiple audio tracks, the text copied to clipboard says "No Audio". Merging nevertheless. I think, I need to clean up propsCopyToClipboard() anyway.

szlldm:
Yes, index at _if((QString)listOfValues.at(5) != QString("00"))_ should have been incremented.

eumagga0x2a:
> Yes, index at _if((QString)listOfValues.at(5) != QString("00"))_ should have been incremented.

I've purged all this insanity I introduced 4 years ago being utterly clueless back then. The dialog should be maintainable now.

PR #308, opened 2021-08-11
[videoFilter] New filter: Zoom

szlldm:
Partializable crop filter.
After release :)


avidemux_plugins/ADM_videoFilters6/zoom/qt4/Q_zoom.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+        connect( ui.horizontalSlider,SIGNAL(valueChanged(int)),this,SLOT(sliderUpdate(int)));
+        connect( ui.pushButtonReset,SIGNAL(clicked(bool)),this,SLOT(reset(bool)));
+#define SPINNER(x) connect( ui.spinBox##x,SIGNAL(valueChanged(int)),this,SLOT(valueChanged(int))); 
+          SPINNER(Left);
+          SPINNER(Right);
+          SPINNER(Top);
+          SPINNER(Bottom);
+
+        show();
```
eumagga0x2a, line 57: I would prefer not to show() a dialog in ctor but handle all necessary proceedings in showEvent() which is then invoked by exec(). At least I would want to understand the reasons why it was necessary to do it differently here.

avidemux_plugins/ADM_videoFilters6/zoom/qt4/Q_zoom.h
```diff
@@ -0,0 +1,37 @@
...
+#include "zoom.h"
+#if 1
+class Ui_zoomWindow : public QDialog
+{
+	Q_OBJECT
+
+protected: 
+	int lock;
+
+public:
```
eumagga0x2a, line 16: ctor and dtor need to be public, sure. But it would be great if the stuff which doesn't need to be accessible from outside the class were kept protected or private.

avidemux_plugins/ADM_videoFilters6/zoom/qt4/Q_zoom.cpp
szlldm: This filter was made with the help of copy-paste-merge technology© :)
This is especially true for the qt4 part, if i remember correctly, i did not change a thing (its from crop filter)

avidemux_plugins/ADM_videoFilters6/zoom/qt4/Q_zoom.h
szlldm: copy-paste-merge technology©
blame the crop filter :)

eumagga0x2a:
> After release :)

Not sure about that. A release in August is not feasible anymore due to my availability constraints. Facing a delay of ~ one month, it might be possible to merge and polish this one in time.

avidemux_plugins/ADM_videoFilters6/zoom/qt4/Q_zoom.cpp
eumagga0x2a: Indeed, I noticed that you kept the qt4 version just the moment I hit "send". The crop filter needs a fix then.

I would not object leaving qt4 out. I didn't try to build Avidemux with Qt4 even once since many months. It is much more important to verify that everything works with qt6.

szlldm:
I wouldn't waste time on QT4 UI, but leave be until it works. The QT5 UI should work with QT6, if Crop filter do.

eumagga0x2a:
Looks great, thanks!

I'll add the include required after the toolkit header split myself.

PR #309, opened 2021-08-12
[imageStab] fix header

szlldm:
just header change


eumagga0x2a:
Thank you for the attribution fix.

PR #313, opened 2021-08-20
[fadeFromImage] simplify direction input; add Push effect

szlldm:
- UI direction input simplified to combobox
+Push effect


eumagga0x2a:
Oops, forgot about this one.

Thanks!

PR #314, opened 2021-08-24
[fadeThrough] generate Fade In and Fade Out from this code base + bugfix

szlldm:
Generate fadeIn and fadeOut from the codebase of the fadeThrough filter, to enjoy the same feature set.


eumagga0x2a:
Thank you, could you please look into arrow buttons in the blur radius spinbox working?

We also probably don't need slider values in a tooltip when the same value is displayed in the corresponding spinbox.

In `ADMVideoFadeThrough::FadeThroughProcess_C` the coefficients for blending are recalculated over and over again – could it be done just once in update()? rgbColorBlend is not going to change between individual pictures, isn't it?

The last question for now:

preview-overlapping-text

This is how the preview opens. Trying to resize the window triggers a massive resize which corrects overlapping parts. Would it be possible to avoid overlap from the very beginning?

szlldm:
> could you please look into arrow buttons in the blur radius spinbox not working?

Indeed, i missed that, thanks!

> In ADMVideoFadeThrough::FadeThroughProcess_C the coefficients for blending are recalculated over and over again – could it be done just once in update()? rgbColorBlend is not going to change between individual pictures, isn't it?

It does change, according to the transient waveform. Even the yuv[3] values could change in theory, if img->_range change (appended segments?).

> Would it be possible to avoid overlap from the very beginning?

I thought setMinimumWidth in showEvent() would do it... Clearly not. Unfortunately i can not replicate this issue on linux.

eumagga0x2a:
>> In ADMVideoFadeThrough::FadeThroughProcess_C the coefficients for blending are recalculated over and over again – could it be done just once in update()? rgbColorBlend is not going to change between individual pictures, isn't it?
> 
> It does change, according to the transient waveform. Even the yuv[3] values could change in theory, if img->_range change (appended segments?).

I mean this part:

```
        // rgb to yuv:
        int rgb[3], yuv[3];
        rgb[0] = (param.rgbColorBlend>>16)&0xFF;
        rgb[1] = (param.rgbColorBlend>>8)&0xFF;
        rgb[2] = (param.rgbColorBlend>>0)&0xFF;
        if(img->_range == ADM_COL_RANGE_MPEG)
        {
            yuv[0] = std::round( 0.257*rgb[0] + 0.504*rgb[1] + 0.098*rgb[2]) + 16;
            yuv[1] = std::round(-0.148*rgb[0] - 0.291*rgb[1] + 0.439*rgb[2]) + 128;
            yuv[2] = std::round( 0.439*rgb[0] - 0.368*rgb[1] - 0.071*rgb[2]) + 128;
            for (int i=0; i<3; i++)
                if (yuv[i] <   16) yuv[i] = 16;
            if (yuv[0] > 235) yuv[0] = 235;
            if (yuv[1] > 240) yuv[1] = 240;
            if (yuv[2] > 240) yuv[2] = 240;
        } else {
            yuv[0] = std::round( 0.299*rgb[0] + 0.587*rgb[1] + 0.114*rgb[2]);
            yuv[1] = std::round(-0.169*rgb[0] - 0.331*rgb[1] + 0.500*rgb[2]) + 128;
            yuv[2] = std::round( 0.500*rgb[0] - 0.419*rgb[1] - 0.081*rgb[2]) + 128;
            for (int i=0; i<3; i++)
            {
                if (yuv[i] <   0) yuv[i] = 0;
                if (yuv[i] > 255) yuv[i] = 255;
            }
        }
        
        // fix swapped uv
        int uvswap = yuv[1]; yuv[1] = yuv[2]; yuv[2]=uvswap;
```

rgbColorBlend in fadeThrough struct is not changing, no? For the case of range changing on-the-fly, we would need to pre-calculate two sets of coefficients, this is true.

> > Would it be possible to avoid overlap from the very beginning?
> 
> I thought setMinimumWidth in showEvent() would do it...

Already found out, it is just a final `adjustSize()` missing, not pushing myself as it could collide with your work. However, as Qt tries to keep window size below 2/3 of available screen real estate, the zoom level of the preview can become tiny and might need some active fightback.

PR #315, opened 2021-08-24
Fadethrough fix

szlldm:
Blur spinbox fix, slider tooltips removed.
Calculate YUV blend parameters only when necessary.

I left out adjustSize(), because it makes the preview super tiny. Maybe You can solve this.


eumagga0x2a:
Great, thanks!

> I left out adjustSize(), because it makes the preview super tiny. Maybe You can solve this.

It looks like we just need to adjustSize() first, setMinimumSize() second.

PR #319, opened 2021-09-02
[waveletSharp] add cutoff parameter

szlldm:
I had an idea, and it looks like it greatly improves the quality of this sharpener.
Increasing the cutoff parameter can totally eliminate sharpening of compression artifacts (blocking) and small noise.


eumagga0x2a:
Thanks!

PR #320, opened 2021-09-03
Exclude end time (marker B) from filtering in case of partial filter, or filter with time scope.

szlldm:


eumagga0x2a:
Thank you very much.

PR #322, opened 2021-09-04
[resampleFps/motin] quality and performance improvements

szlldm:
less artifacts
30-40% faster (at least on an 8 core+HT machine)


eumagga0x2a:
Thanks!

PR #325, opened 2021-09-16
[flyDialog] maximize

szlldm:
Add maximize button and open flyDialog maximized _only_ if the main window is maximized.
It is added conditionally, because it requires to change dialog mode to window mode.

There is a bug, unmaximizing the filter window, the preview scaling will be bogus. Manually resizing the window the issue vanishes.


eumagga0x2a:
Thanks, with this patch the build on macOS fails:

```
Undefined symbols for architecture x86_64:
  "_QuiMainWindows", referenced from:
      ADM_flyDialog::ADM_flyDialog(QDialog*, unsigned int, unsigned int, ADM_coreVideoFilter*, ADM_QCanvas*, ADM_QSlider*, ResizeMethod) in DIA_flyDialog.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [ADM_UIsQt4/src/libADM_UIQT66.dylib] Error 1
make[1]: *** [ADM_UIsQt4/src/CMakeFiles/ADM_UIQT66.dir/all] Error 2
make: *** [all] Error 2
** Failed at make**
```

I didn't have a chance to check on other platforms yet. By the way, the preview dialog on macOS had the maximize button all the time, independently from the main window state, however.

Is there a use case for being able to maximize filter preview dialog? It doesn't address the actual problem of the initial dialog size being too small with some filters, does it?

szlldm:
Damn. It builds fine on Linux with
`extern QWidget *QuiMainWindows;`
There is no maximize button on Linux or Windows.
The use case is to view as much details as possible. Very important for example in case of sharpeners.

eumagga0x2a:
While it should be possible to avoid referencing QuiMainWindows be walking up the family tree like

```
    QWidget *top = parent->parentWidget();
    while(top && top->objectName() != "MainWindow")
    {
        top = top->parentWidget();
    }
    if(top && top->objectName() == "MainWindow" && (top->windowState() & Qt::WindowMaximized))
    {
        parent->setWindowFlags(Qt::Window);
        parent->setWindowFlag(Qt::WindowMinimizeButtonHint, false);
        parent->setWindowFlag(Qt::WindowMaximizeButtonHint, true);
        parent->setWindowState(Qt::WindowMaximized);
    }

```

the patch results in the filter preview dialog always opening maximized if the main window is maximized, even if it was normalized in the meantime, very annoying if not desired.

> The use case is to view as much details as possible. Very important for example in case of sharpeners.

Shouldn't this feature be filter-specific then? I am not sure how Qt behaves WRT maximize button for modal dialogs on Windows (need to check it first), but for most filters I would prefer not to have this behaviour.

szlldm:
To cut the gordian knot, filters are maximized if preference is set.

eumagga0x2a:
> the preview dialog on macOS had the maximize button all the time

No wonder, I am the culprit: https://github.com/mean00/avidemux2/commit/c5984942bc5482206ed79ebf38339898d5311766

Last night, I could test a little on Linux. Reflagging dialog window as a normal window results in a very weird behaviour of gnome-shell 3.38.5 / mutter: such a filter preview window can be moved to another workspace, but switching workspaces back becomes impossible(?!) unless preview is closed. Will test with the current gnome-shell and Qt6 later.

Definitely not going the path of adding a preference, QSetting would be fine because filter-specific, but the problem are window flags and how Qt and window managers handle this, else I would suggest the private patch below (GitHub doesn't accept .patch, great, thus renamed to .patch.txt).

[video-filter-preview-persistent-maximize.patch.txt](https://github.com/mean00/avidemux2/files/7189353/video-filter-preview-persistent-maximize.patch.txt)

szlldm:
> Definitely not going the path of adding a preference, QSetting would be fine because filter-specific, but the problem are window flags and how Qt and window managers handle this

How about a preference to open filters as Qt::dialog or Qt::window? Then Qsetting could could restore size if pref set to Qt::window.

eumagga0x2a:
> How about a preference to open filters as Qt::dialog or Qt::window?

I don't think the feature is significant enough to justify introduction of broken behaviour* which unfortunately ensues at least with gnome-shell, the current version included, when moving modal windows not flagged as a dialog to a different workspace.

*) It is impossible to switch to the workspace containing Avidemux main window once filter preview has been moved to a different workspace as long as the filter preview remains open on a different workspace.

> Then Qsetting could could restore size if pref set to `Qt::window`.

I haven't found a way to restore the parent widget size / window geometry from `ADM_flyDialog`, that turned out to be tricky. Restoring window state was easy.

By the way, the part of `qtRegisterDialog` which should ensure that all windows are open as dialogs is buggy, it should have been

```
        if ((flags & Qt::Dialog) == Qt::Dialog)
            isDialog = true;
```
This bug was the reason why the proposed patch was working in the first place ;-D

szlldm:
> I don't think the feature is significant enough to justify introduction of broken behaviour

You are right, this Qt:window mode has some culprits. But the feature is significant enough (at least for me, i hate to manually resize filters every time), so i changed to only maximize the filter  (when preference is set) without changing windowFlags.

PR #331, opened 2021-10-11
Add riscv64 support

Xeonacid:
Tested on Arch Linux RISC-V.


Xeonacid:
@mean00 I think it should, but I have no RV32 environment to test it. So I named it `riscv64` conservatively.
Should I open another PR to rename all `riscv64` occurrences to `riscv`?

mean00:
it's minor as most if not all the linux riscv systems are 64 bits
but defining both ADM_CPU_RISCV and ADM_CPU_RISCV_64 would keep it consistent with their x86 counterpart
Thanks

mean00:
(No need to check for actual xlen==64)

PR #341, opened 2021-10-20
[hdr] config HDR tone mapping from dedicated dialog

szlldm:
HDR tone mapping config (method, saturation, boost) available from the "Video" menu.
In Preferences only default method and target luminance (which could even be a hidden hard coded constant, but keep anyway) are kept.
The config is written out to the python file(s) as "adm.setHDRConfig", so it also can be used in batch processing.

Only **missing** part is the tinyPy readback of "adm.setHDRConfig".


szlldm:
Currently loading python file (for example lastEdit.py) throw a non fatal error, because "adm.setHDRConfig" is unknown.
I was unsure how to add it to tinyPy.

PR #349, opened 2021-10-26
[gui_blackframes] fast backward search

szlldm:
go to the prev key frame, check forward up to the current position, if not found set the keyframe as the current position, then repeat


eumagga0x2a:
I'll apply the fix for the all-intra special case myself.

Thank you very much.

PR #352, opened 2021-10-28
[ADM_flyDialog] sanitizing, fix memleak; [colorBalance/qt4] simplify code

szlldm:


eumagga0x2a:
Thanks, what is the use case for `ControlOption::UserWidgetBeforePeekBtn`?

Is it meant to override `userWidget != NULL` passed to `addControl()`?

If yes, then we get a memleak. If no, it would be enough to check that `userWidget != NULL`.

Do you plan to add an additional button or a different widget (always a single one) to controls in filter preview dialogs other than colorBalance?

eumagga0x2a:
Okay, thank you, this way the changes get a justification. At a cursory glance, it doesn't seem like `ControlOption::UserWidgetAfterPeekBtn` is handled when populating buttonList, could you please check?

eumagga0x2a:
Thank you. I didn't have a chance to test that everything is OK on macOS yet, but I hope it will.

PR #354, opened 2021-11-02
[hdr] show changes

szlldm:
use nextPicture()+previousKeyFrame() to flush cache, and display the effect of the new hdr settings on the closest (prev or same) key frame.

### DIFF
--- a/avidemux/common/ADM_commonUI/DIA_HDRConfig.cpp
+++ b/avidemux/common/ADM_commonUI/DIA_HDRConfig.cpp
@@ -13,7 +13,7 @@
 
 int DIA_getHDRParams( uint32_t * toneMappingMethod, float * saturationAdjust, float * boostAdjust)
 {
-    diaElemReadOnlyText applicability(NULL,QT_TRANSLATE_NOOP("adm","The options above are not immediately effective on cached and displayed frames"),NULL);
+    //diaElemReadOnlyText applicability(NULL,QT_TRANSLATE_NOOP("adm","The options above are not immediately effective on cached and displayed frames"),NULL);
     
     diaMenuEntry toneMapEntries[]={
                           {0,       QT_TRANSLATE_NOOP("adm","disabled"),NULL}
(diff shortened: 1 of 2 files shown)


eumagga0x2a:
Thank you, in case the currently displayed picture is not a keyframe, why did you decide against preserving the position? Because it can take quite a long time at high resolutions and long GOPs?

eumagga0x2a:
I meant something along the lines of

```
diff --git a/avidemux/common/gui_main.cpp b/avidemux/common/gui_main.cpp
index d88a9209b..4a792ef4d 100644
--- a/avidemux/common/gui_main.cpp
+++ b/avidemux/common/gui_main.cpp
@@ -1504,6 +1504,25 @@ void    A_setHDRConfig( void )
      if(DIA_getHDRParams( &method, &saturation,&boost))
      {
         video_body->setHDRConfig(method,saturation,boost);
+        ADMImage *pic = admPreview::getBuffer();
+        uint64_t time = admPreview::getCurrentPts();
+        if(pic)
+        {
+            admPreview::deferDisplay(true);
+            if(pic->flags & AVI_KEY_FRAME)
+            {
+                admPreview::nextPicture();
+                admPreview::previousKeyFrame();
+            }else
+            {
+                admPreview::previousKeyFrame(); // flush cache
+                if(false == admPreview::seekToTime(time))
+                    video_body->rewind(); // avoid crash
+            }
+            admPreview::deferDisplay(false);
+            admPreview::samePicture();
+            GUI_setCurrentFrameAndTime();
+        }
      }
 
 }

```

szlldm:
Yes, it go to keyframe, not just because decoding can take time, but also tonemapping needs to be done on decoded frames up to the current position. And top of that, there is no feedback (busy cursor).

eumagga0x2a:
So missing busy cursor was the reason not to seek to the current position, if I got you correctly? Why not adding the ACT_SetHDRConfig to `MainWindow::actionSlot(Action)` to get one?

szlldm:
There would be a busy cursor during config dialog.
But the real reason is speed. With long GOP and rgb mapper it is an eternity.

eumagga0x2a:
> There would be a busy cursor during config dialog.

Indeed, just realized.

> But the real reason is speed. With long GOP and rgb mapper it is an eternity.

This is true, but from UX, being thrown out of context by seeking to a different location might be even worse. In any case, I suspect we would need `GUI_setCurrentFrameAndTime();` to sync time display etc. if we end up at the previous keyframe.

eumagga0x2a:
> There would be a busy cursor during config dialog.

Obviously, this affects ACT_GotoTime too. A fix via a wrapper action and a global variable might be possible but not exactly pretty.

eumagga0x2a:
I have a solution for showing busy cursor during seek / refresh only, after the dialog was closed. Do you steadfastly object keeping the position if not on a keyframe?

szlldm:
with busy indicaton its maybe ok

eumagga0x2a:
Good, I'll merge this and add the action split, which is needed for ACT_gotoTime anyway.

PR #355, opened 2021-11-02
[ADM_flyDialog] add analyzer option

szlldm:
preliminary analyzer option
main issue is resize
for testing added to "Color balance" filter

### DIFF
--- a/avidemux/qt4/ADM_UIs/include/DIA_flyDialogQt4.h
+++ b/avidemux/qt4/ADM_UIs/include/DIA_flyDialogQt4.h
@@ -33,6 +33,8 @@
 #include <QMouseEvent>
 #include <QPoint>
 #include <QRect>
+#include <QGraphicsScene>
+#include <QImage>
 
 #include "ADM_default.h"
 #include "ADM_rgb.h"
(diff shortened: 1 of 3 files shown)


eumagga0x2a:
Truly amazing, works great on Linux, but FYI, on macOS the analyzer window doesn't stay in foreground when user interacts with the parent filter preview dialog. The special handling of dialogs on macOS in qtRegisterDialog() seems to be redundant now, I'll remove it, but it is not the culprit here, disabling it doesn't help.

On macOS, it is also necessary to manually resize the analyzer window to make sceneVectorScope fit in. When resizing, the background on one or both sides of the graphs is not painted, remains white.

Do you plan to add some guidance for users? Also the size of the graphs is so tiny that while looking very impressive, it is hard to actually make use of the information. Maybe lay out it as a set of tabs?

Apart from this, ahem, I'd suggest renaming flyDialogsAnalyzer member of ADM_flyDialog to avoid attracting ridicule. Yeah, it is the name of the branch :-D

szlldm:
Its a blind try to fix dialog going into background on macOS.
I have no idea how to fix the other issues.
A will try tabbed layout, or adding checkboxes (also would serves as label).

eumagga0x2a:
> Its a blind try to fix dialog going into background on macOS.

It hasn't helped, unfortunately.

In general, ordering of modal windows is a mess on macOS. Using Mission Control results in the filter manager dialog becoming the top window above open filter preview while remaining disabled.

I think, targeting 2.8.1+ with this work would remove the pressure to deliver a polished result. It is a pity that we keep releasing from git master rather than a dedicated release branch.

PR #356, opened 2021-11-05
[GUI] CTRL-key modifier for main navigation slider

szlldm:
CTRL + wheel -> goto next/prev. frame
CTRL + click/drag -> goto nearest frame


eumagga0x2a:
Fantastic, works great with the navigation slider too, thank you very much.

PR #357, opened 2021-11-05
[ADM_preview] omit error message if can not go to the next/prev keyframe, but there is none to go to

szlldm:
i find annoying, when it pops up error message, while at the beginning or at the end of the video

### DIFF
--- a/avidemux/common/ADM_preview.h
+++ b/avidemux/common/ADM_preview.h
@@ -39,8 +39,8 @@ class admPreview
       static void cleanUp(void);
       static ADMImage *getBuffer(void);
       static uint64_t getCurrentPts(void);
-      static bool nextKeyFrame(void);
-      static bool previousKeyFrame(void);
+      static bool nextKeyFrame(bool * keyframeNotFound = NULL);
+      static bool previousKeyFrame(bool * keyframeNotFound = NULL);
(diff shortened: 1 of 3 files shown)


eumagga0x2a:
Please imagine a somewhat damaged video with a keyframe missing a timestamp so that `getNKFramePTS()` fails. Alternatively, it might be also a video completely without keyframes after the first one (such a stream may be impractical, but otherwise entirely valid). How should we let the user know that she really cannot seek to the next keyframe and hammering the up or down arrow key won't help?

Another, much more frequent case might be a relatively short video with very long GOPs. The user sees that there is still a significant distance before the end of the navigation slider range, but she cannot advance to the next keyframe. A reason to get mad? Sure!

I don't want to defend the current error message as it conveys zero information about the background of the seek failure, but it at least provides assurance that further attempts to go to the next keyframe will be futile.

szlldm:
I understand the reasoning. But the current solution is also maddening. The popup error message is annoying not just because it interrupts the editing process, but suddenly the user can not discern that it means their video is damaged, or just reached the end/begining. (In case of undecodable key frame you get the error, can't go through with nextkeyframe, but can be jumped over with the navigation slider.)
Maybe add a user interface preference? Or extend the Message level preference?

PR #358, opened 2021-11-07
[GUI] add hidden shortcuts to Go menu; fix swap up/down in case of 1 min jump

szlldm:
-add hidden shortcuts (Back/Forth 1,2,4 sec & 1 min) to the Go menu, let the users know about them
-KEYBOARD_SHORTCUTS_SWAP_UP_DOWN_KEYS was no applied on Back/Forth 1 min shortcut.


eumagga0x2a:
Thanks for tackling this and catching the bug!

It feels also like a good choice to be consistent with the navigation widget regarding spelling out the number one in "one minute" vs using numerals in 1/2/4 seconds-based navigation.

PR #359, opened 2021-11-09
[GUI] enable some navigation during play

szlldm:
enable backward/forward time jumps (1,2,4,60sec) and gotomarkers.


szlldm:
also during play fast seeking possible via left & right keys (however the corresponding next/prev frame buttons are not activated, to avoid confuse)

eumagga0x2a:
Works great, thanks, just the choice to use left/right to seek to nearby keyframes while up/down remain disabled feels now somewhat strange.

PR #364, opened 2021-11-11
[filtermanager] improvements

szlldm:


eumagga0x2a:
Great, allowing to revert making a filter partial closes a substantial gap in the functionality, thank you very much.  But could you please suggest another term (at least for the GUI) for extracting the filter instance swallowed by the partial filter? In my perception, partial filter hasn't sinned ;-)

It might be also just a question of personal taste, but maybe ADM_vf_partialize(int) could handle both partializing and globalizing the filter, i.e. turning into a toggle.

Regarding left/right key navigation: a great idea as long as we don't need one of the lists to be horizontally scrollable. Currently we don't, later users won't let us take this feature away, I fear :-)

szlldm:
>  In my perception, partial filter hasn't sinned

[https://wikidiff.com/absolve/absolution](url)

> maybe ADM_vf_partialize(int) could handle both partializing and globalizing the filter

IMHO its more clear&clean if separated.

eumagga0x2a:
Thanks!

PR #366, opened 2021-11-15
[eq2] fix green tint

szlldm:
there was a slight, but noticeable green tint even with default settings.


eumagga0x2a:
I remember this having been reported, thank you for the fix.

PR #371, opened 2021-11-19
[GUI] very quick method to add partialized filters

szlldm:
-using the Ctrl+F shortcut (or through Video menu) we get a dialog with the list of partializable filters.
-filters can be selected by regular navigation (arrow keys, scrolling, etc) or by starting typing their names (like press "z" for Zoom)
-closing (accepting) the selected filter's configuration dialog the filter list dialog also closes, saving one click/key press


eumagga0x2a:
Thank you very much. Further refinements can be done in-tree.

PR #373, opened 2021-12-01
Fix typos

adem4ik:


eumagga0x2a:
Could you please submit a PR for avidemux2_i18n manually adjusting all translation files which provide mostly correct translations for defective source strings fixed here? I would like to avoid invalidating them.

This is actually the reason why such typos live on – it is pretty cumbersome to fix source strings in translation files without losing translations.

adem4ik:
@eumagga0x2a Got it, I'll do it later this week.

eumagga0x2a:
My apologies, I cancel my request, all necessary editing of translation files done instantly using two sed one-liners.

Thank you.

PR #374, opened 2021-12-01
[ADM_videoFilters] bugfix: recreate filter chain if partial filter reconfigured

szlldm:
The preview of a last partialized Add logo filter had messed up timing if their parent partial filter range changed.


eumagga0x2a:
Thanks!

PR #378, opened 2021-12-04
[gui_main] prevent crash when restoring previous session with deleted video

szlldm:
accidentaly tried to restore a previous session, while the source file was already deleted. this led to a crash


eumagga0x2a:
Thanks for catching this.

PR #379, opened 2021-12-04
[ADM_segment] fix crash when loading keyframeless/bogus video while shared cache enabled

szlldm:
with enabled shared cache it was crashing when trying to load video in
[https://avidemux.org/smif/index.php?msg=92917](https://avidemux.org/smif/index.php?msg=92917)


eumagga0x2a:
Oh yes, thanks.

PR #382, opened 2021-12-08
[ADM_coreImage] performance improvements

szlldm:


avidemux_core/ADM_coreImage/src/ADM_image.cpp
```diff
@@ -179,13 +179,14 @@ bool BitBlitAlpha(uint8_t *dst, uint32_t pitchDst,uint8_t *src,uint32_t pitchSrc
  */
 bool BitBlit(uint8_t *dst, uint32_t pitchDst,uint8_t *src,uint32_t pitchSrc,uint32_t width, uint32_t height)
 {
-#if 1
+#if 0
+    // av_image_copy_plane is just a wrapper around a wrapper of line-by-line memcpy
```
eumagga0x2a, line 184: The comment from https://github.com/mean00/avidemux2/commit/6b5b805826a603935742de5c2c838e2ac05dba63 points to av_image_copy_plane wrapping SSE4 `image_copy_plane_uc_from` from libavutil/x86/imgutils.asm. Are you sure it is not faster than a plain memcpy, given the CPU supports SSE4 and Avidemux doesn't tell FFmpeg not to use it?

szlldm: av_image_copy_plane calls image_copy_plane (without the uc_from extension, which is used for unCached memory)

static void image_copy_plane(uint8_t       *dst, ptrdiff_t dst_linesize,
                             const uint8_t *src, ptrdiff_t src_linesize,
                             ptrdiff_t bytewidth, int height)
{
    if (!dst || !src)
        return;
    av_assert0(FFABS(src_linesize) >= bytewidth);
    av_assert0(FFABS(dst_linesize) >= bytewidth);
    for (;height > 0; height--) {
        memcpy(dst, src, bytewidth);
        dst += dst_linesize;
        src += src_linesize;
    }
}

avidemux_core/ADM_coreImage/src/ADM_imageOperation.cpp
```diff
@@ -60,7 +60,10 @@ bool ADMImage::duplicateMacro(ADMImage *src,bool swap)
                         }
                     }
                 }
-                BitBlit(dest, destStride,source,sourceStride,opWidth, opHeight);
+                if (destStride == sourceStride)
+                    memcpy(dest, source, destStride*opHeight);  // copying in one run is faster then line-by-line
```
eumagga0x2a, line 64: If a single memcpy call wins over an SSE4-accelerated implementation in libavtuil (it certainly does on non-x86, so ARM users should rejoyce), wouldn't it be appropriate to handle this special case in BitBlit itself rather than letting callers deal with it individually?

szlldm: there is NO SSE4 accelerated stuff in libavutil

szlldm: BitBlit can not use a single memcpy, becuase it doesn't know important parameters

avidemux_core/ADM_coreImage/src/ADM_image.cpp
eumagga0x2a: Oops, indeed, now I wonder whether the same reading glitch happened to the author of https://github.com/mean00/avidemux2/commit/6b5b805826a603935742de5c2c838e2ac05dba63 ...

The uc_from implementation should be the one which was hoped to help with slow download speed from VA-API hw surfaces, if I am not mistaken.

szlldm: yes

avidemux_core/ADM_coreImage/src/ADM_imageOperation.cpp
eumagga0x2a: > there is NO SSE4 accelerated stuff in libavutil

Purely my reading glitch, sorry.

> BitBlit can not use a single memcpy, becuase it doesn't know important parameters

If the caller uses memory beyond the width of the picture but before the end of the stride for some purposes, the operation will be destructive, yes. Is the assumption safe that this cannot happen in `ADMImage::duplicateMacro`?

En-bloc copy evaluates only that source and destination strides match. Have I missed something else?

szlldm: BitBlit is used in other scenarios, similar to a "Picture-in-picture" function, where a full copy would destroy visible pixels.
In ADMImage::duplicateMacro the width and height match is asserted, only the stride remains, which is checked before memcpy. I dont see a possibility (copying same size ADM_images) where the BitBlit would be used instead of memcpy, but left there just in case.

eumagga0x2a: Ah, indeed, in `ADMImage::copyTo`, thank you for pointing out.

avidemux_core/ADM_coreImage/src/ADM_image.cpp
```diff
@@ -152,15 +152,16 @@ bool BitBlitAlpha(uint8_t *dst, uint32_t pitchDst,uint8_t *src,uint32_t pitchSrc
     if(alpha>255) alpha=255;
     if(alpha==255) // take a shortcut
         return BitBlit(dst,pitchDst,src,pitchSrc,width,height);
+    float af = alpha * (256.0/255.0);
```
eumagga0x2a, line 155: I'm not sure that this won't collapse into `float af = alpha * 1;` with clang. The latter may require explicit cast to float for the first operand.

avidemux_core/ADM_coreImage/src/ADM_image.cpp
```diff
@@ -152,15 +152,16 @@ bool BitBlitAlpha(uint8_t *dst, uint32_t pitchDst,uint8_t *src,uint32_t pitchSrc
...
     {
         for(int x=0;x<width;x++)
         {
-            float s=src[x],d=dst[x];
-            d=s*alpha+(255.-alpha)*d;
-            d/=255.;
-            d+=0.49;
-            dst[x]=(uint8_t)d;
+            pix = alpha*src[x] + negalpha*dst[x];
+            dst[x]=(uint8_t)(pix>>8);
```
eumagga0x2a, line 164: I really wished there were a proper issue tracker as I absolutely cannot tell now what test case was the reason for https://github.com/mean00/avidemux2/commit/6dac2d0d31d9e63d9bf3412f0633fbbe04b5e889 this commit basically reverts.

avidemux_core/ADM_coreImage/src/ADM_image.cpp
szlldm: i hope not, otherwise clang would be an unusable compiler :-)

avidemux_core/ADM_coreImage/src/ADM_image.cpp
szlldm: imho rounding errors wouldnt produce any noticable effects.

avidemux_core/ADM_coreImage/src/ADM_image.cpp
szlldm: it would violate basic Implicit type conversion rules

eumagga0x2a: Implicit conversion in the above statement verified working with clang :-)

eumagga0x2a: Interestingly, now I can revert https://github.com/mean00/avidemux2/commit/a29b29b35ee9c0b0a4be7383b7a1a05414eda400 and fadeToBlack still works fine on Monterey. I cannot rule out that Apple clang was indeed broken a while ago, but has been fixed since then. I just cannot verify that this extends to the version available for Catalina as well.

eumagga0x2a:
Haven't found any issues with integer BitBlitAlpha. We'll need at least one more round of nightly builds, though.

Thank you.

PR #381, opened 2021-12-07
[ADM_threadQueue] properly threaded producer-consumer pipeline

szlldm:

PR #383, opened 2021-12-10
[ADM_threadQueue] bugfix + exclusive access for state variable

szlldm:
-fix stopping issue  (half second delay) during playback (which resulted very slow navigation during play)
-always access threadState exclusively

### DIFF
--- a/avidemux/common/ADM_audioFilter/src/audiofilter_thread.cpp
+++ b/avidemux/common/ADM_audioFilter/src/audiofilter_thread.cpp
@@ -122,6 +122,11 @@ bool    ADM_audioAccess_thread::getPacket(uint8_t *buffer, uint32_t *size, uint3
     while(1)
     {
         mutex->lock();
+        if(threadState==RunStateStopOrder)  
+        {
+            mutex->unlock();
+            return false;
+        }
         if(list.size())
         {
             //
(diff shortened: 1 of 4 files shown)


mean00 (commented):
Some of the new locks are not needed IMHO.
The scheme you want to avoid is read/modify/write

something like this : 
lock
threadState=running
unlock
does not help i think, unless threadstate=running is not atomic

Same for the read only/comparison only test

szlldm:
"According to the C++11 ISO Standard, the volatile keyword is only meant for use for hardware access; do not use it for inter-thread communication."

mean00:
Let me detail, If the variable is of simple type (i.e. int/enum) and fits in a register

foo=bar

and 

lock()
foo=bar
unlock()

are identical as far as protection are concerned
The 2nd may be much slower

szlldm:
Quoting from the top answer at [https://stackoverflow.com/questions/2484980/why-is-volatile-not-considered-useful-in-multithreaded-c-or-c-programming](https://stackoverflow.com/questions/2484980/why-is-volatile-not-considered-useful-in-multithreaded-c-or-c-programming).
"volatile does not guarantee that no reordering takes place. Assume that we use a volatile variable as a flag to indicate whether or not some data is ready to be read. In our code, we simply set the flag after preparing the data, so all looks fine. But what if the instructions are reordered so the flag is set first?"

szlldm:
std::atomic could be used, but that would limit to C++11 and later

szlldm:
+1: volatile != atomic. there could be some architecture, where writing a volatile address, would lead to a read-modify-write (for example "volatile uint8_t", on some 32bit arch)

eumagga0x2a:
Merging now. In case of substantial regressions which cannot be solved quickly I'll back out this one and #381, but test builds with both included run fine so far.

PR #384, opened 2021-12-15
[ADM_imageSave] multithreading saveAsJpgInternal

szlldm:
Selection save can be sped up


eumagga0x2a:
The effect is not very pronounced on my hardware (I got ~10% speedup), but it is an improvement. Thanks!

PR #388, opened 2021-12-29
[prefs] Option to reverse mouse wheel for navigation

szlldm:


eumagga0x2a:
Shouldn't the new option cover filter preview navigation as well?

szlldm:
> Shouldn't the new option cover filter preview navigation as well?

Didnt even noticed it can be scrolled :) Hovewer, neither "Reverse UP and DOWN arrow keys" cover filter preview.

eumagga0x2a:
> Hovewer, neither "Reverse UP and DOWN arrow keys" cover filter preview.

In many filter configuration dialogs, catching arrow keys might be detrimental to usability.

szlldm:
> In many filter configuration dialogs, catching arrow keys might be detrimental to usability.

If slider is in focus, arrow keys work on that.
In general, scrolling (and arrow keys as well) in filter previev is unusable (it doesnt jump to a keyframe, but increment the hidden value), especially with longer GOPs.
Definetly wont be solved in this PR. Maybe later.

szlldm:
First, flyDialogs should have repalce simple navigation sliders with similar to the one in the main window (so as bare minimum support goto keyframes, but goto cutpoint would be also very useful).
If this is done (it is on my TODO list, but not on the top), only then would be meaningful to apply this preference in filter's preview IMVHO.
Until then this (separated) PR is limited on the main window's navigation slider, but it can be very useful on its own, as scrolling down can be more natural&comfortable, and therefore there should be an option to use that direction for forward navigation.

szlldm:
Now the new option covers both filter preview navigation, as well filter manager's preview.
Additionally the navigation slider of the previews show the markers. In case of partialized filters, this means the time scope of the partialization.

szlldm:
rebased to current master (then force-pushed), to cover cubicLUT filter

eumagga0x2a:
Thank you very much for this arduous work.

> First, flyDialogs should have repalce simple navigation sliders with similar to the one in the main window (so as bare minimum support goto keyframes

I doubt that this fits into design concept. Filters are unaware of any underlying complexity related to the internal structure of video stream(s) which is fully encapsulated by the editor. I really would prefer not to leak it in any way whatsoever.

PR #389, opened 2021-12-30
[gui_save] Save selection as PNG

szlldm:
also set PNG compression level to minimal (set to one, as zero would be uncompressed)
this results aprox. 5% larger files, but saving is 30-40% faster


szlldm:
enabling "paeth" prediction, file size can be ~ 30% smaller, while the encoding is 15-20% faster

PR #390, opened 2021-12-30
[prefs] performance improvements

szlldm:
avoid full search (especially by name) during lookup if possible


eumagga0x2a:
I may have easily missed another location, but it looks like after you thankfully fixed the storm of pointless queries in https://github.com/mean00/avidemux2/pull/391 the only location where we stress preferences is in `ADMToneMapperConfig::getConfig`, which should be trivial to fix at the source, so that prioritizing simplicity over performance won't hurt.

eumagga0x2a:
Frequent polling in tone mapper should be resolved by https://github.com/mean00/avidemux2/commit/26490fa84e110fcb712a15ec233708ab1d5b6afe but please keep this PR open, at least for a while.

PR #391, opened 2021-12-30
[Q_gui2] event handler only query preference if relevant

szlldm:
for example event handler fires rapidly if mouse cursor moving over the main window


eumagga0x2a:
A great catch, thanks!

PR #396, opened 2022-01-08
[GUI] add themes

szlldm:
user can change themes on the fly (through View/Themes menu):
-Native theme (option not shown if native theme is fusion)
-Light theme (fusion with default colors)
-Dark theme (fusion with dark colors)

saved to QSettings

### DIFF
--- a/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
+++ b/avidemux/qt4/ADM_userInterfaces/ADM_gui/Q_gui2.cpp
@@ -20,6 +20,8 @@
 #include <QGraphicsView>
 #include <QtCore/QDir>
 #include <QMessageBox>
+#include <QPalette>
+#include <QColor>
 #if QT_VERSION < QT_VERSION_CHECK(5,11,0)
 #   include <QDesktopWidget>
 #else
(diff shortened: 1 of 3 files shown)


eumagga0x2a:
Probably a minor inconvenience seen at least with Qt 6.2.2, but it might still matter when this switch is meant to be on-the-fly:

menu-text-color-after-switching-from-dark-to-light

STR:
1. Switch to dark theme (fusion with custom palette)
2. Restart Avidemux
3. Switch to light theme, don't restart Avidemux yet
4. Open drop-down menu of any of the combo boxes in the codec widget

The same happens on a switch from light to dark, just that the barely visible text is dark on dark. The text in the time indicator has wrong color too.

Restarting Avidemux rectifies that.

I haven't tried yet whether this will work on-the-fly at all, but I personally don't like the idea of allowing users to switch Qt theme away from the native one on non-Linux platforms. If only Apple didn't break native Qt theme on almost every macOS upgrade...

szlldm:
i try to look into the issue.

> but I personally don't like the idea of allowing users to switch Qt theme away from the native one

this should be the user's decision, right? the default theme anyway is the native one.

szlldm:
Can not replicate the issue with Qt 5.12.8.
Its either a bug in 6.2.2, or there are some new member in Qt 6, i obviously didnt used. On the other hand, if restarting helps, it makes the bug case more probable.

eumagga0x2a:
> Can not replicate the issue with Qt 5.12.8.

Cannot reproduce `QPalette::WindowText` from new palette not applied with Qt 5.15.2 either. I agree that this looks like a bug in Qt6.

The thick red warning below the description of `QStyle *QApplication::setStyle(const QString &style)` both for Qt5 and Qt6 hints at a limitation acknowledged and tolerated by Qt developers.

> this should be the user's decision, right?

For those who are advanced enough to set QT_STYLE_OVERRIDE and who are aware of the existence of different Qt themes, this is okay. But I expect people thinking that the dark mode is somewhat broken if it doesn't look native (and of course because the window frame remains bright no matter what).

szlldm:
> But I expect people thinking that the dark mode is somewhat broken if it doesn't look native

that is the reason there is a Light theme besides Native :)

eumagga0x2a:
Thank you. I'll probably need to change the way the menuThemes submenu is added in order to make it easier to disable it for specific platforms, but this doesn't need to happen right now.

PR #397, opened 2022-01-08
[ADM_demuxers/Matroska] reuse indexing dialog

szlldm:
prevent very annoying focus stealing, by reusing the same dialog for cluster and image indexing
(wondering if cluster and image indexing could be united into a single process, as after a cluster found image the images in it. this would be faster in case file located on a HDD)

also prevent memory leak (undeleted work)

### DIFF
--- a/avidemux/cli/ADM_userInterfaces/ADM_dialog/DIA_working.cpp
+++ b/avidemux/cli/ADM_userInterfaces/ADM_dialog/DIA_working.cpp
@@ -39,6 +39,7 @@ public:
     virtual uint8_t     update(uint32_t percent);
     virtual uint8_t     update(uint32_t current,uint32_t total);
     virtual uint8_t     isAlive (void );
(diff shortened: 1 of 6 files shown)


eumagga0x2a:
> wondering if cluster and image indexing could be united into a single process, as after a cluster found image the images in it. this would be faster in case file located on a HDD

Exactly this. Solves the issue of focus stealing at its source.

szlldm:
So 92b9d49b76588c4812f3aaf5c06dd973594e4eff made this irrelevant? :)

eumagga0x2a:
> So [92b9d49](https://github.com/mean00/avidemux2/commit/92b9d49b76588c4812f3aaf5c06dd973594e4eff) made this irrelevant? :)

For the purpose of preventing Avidemux window moved to a different workspace from suddenly popping up in foreground on the current one in the middle of indexing a Matroska video – yes.

szlldm:
👍

PR #398, opened 2022-01-20
[hue] make partializable

szlldm:


eumagga0x2a:
I wonder why hue as well as mSharpen were implemented using filter cache back then. Especially like here, with cache size of just one.

szlldm:
> I wonder why hue as well as mSharpen were implemented using filter cache back then. Especially like here, with cache size of just one.

¯\_(ツ)_/¯

eumagga0x2a:
Thank you.

PR #400, opened 2022-01-26
[ADM_toneMapper/RGB] bugfixes and performance improvements

szlldm:
-much better/accurate colors
-significantly faster (almost x2 faster on many core CPU)


eumagga0x2a:
Thank you.

Regarding your comment

> // when multithreaded libswscale available retest with AV_PIX_FMT_GBRP16LE

– could you please elaborate?

szlldm:
ffmpeg 5.0 has multithreaded libswscale. so when it will get incorporated into adm, then it would be advised to test, if using AV_PIX_FMT_GBRP16LE would be faster, than the makeshift multithreaded conversion

eumagga0x2a:
I see, thanks.

PR #404, opened 2022-02-04
PostProcessing improvements

szlldm:
-show postporc settings changes in the shown frame (similar as tonemapping)
-separate deblocking parameters from deinterlacing (strength does'nt affect deinterlacing)
-make deblocking functional


eumagga0x2a:
Yes, this absolutely makes sense to handle deinterlacing differently, thanks!

PR #405, opened 2022-02-05
Partial support for videos with variable resolution

szlldm:
-works* with software decoderFF
-vdpau fail* with "malloc(): corrupted top size" @ hwDownloadFromRef()  --> HW ref image handling commented out

* only could test with a single mpeg2 file, aom, vpx untested


szlldm:
it works with libva, can't test dxva
vdpau&dxva commented out @ ADM_edRenderInternal.cpp:line439
in case of unsupported hw decoder instead of crashing it shows blackened image with conflicting resolution string

PR #407, opened 2022-02-15
[hdr] adaptive RGB tonemappers

szlldm:
+fix limited range YCbCr to RGB conversion

### DIFF
--- a/avidemux/common/ADM_commonUI/DIA_HDRConfig.cpp
+++ b/avidemux/common/ADM_commonUI/DIA_HDRConfig.cpp
@@ -25,6 +25,9 @@ int DIA_getHDRParams( uint32_t * toneMappingMethod, float * saturationAdjust, fl
                          ,{2,       QT_TRANSLATE_NOOP("adm","RGB clipping"),NULL}
                          ,{3,       QT_TRANSLATE_NOOP("adm","RGB Reinhard"),NULL}
                          ,{4,       QT_TRANSLATE_NOOP("adm","RGB Hable"),NULL}
+                         ,{5,       QT_TRANSLATE_NOOP("adm","RGB clipping - adaptive"),NULL}
+                         ,{6,       QT_TRANSLATE_NOOP("adm","RGB Reinhard - adaptive"),NULL}
+                         ,{7,       QT_TRANSLATE_NOOP("adm","RGB Hable - adaptive"),NULL}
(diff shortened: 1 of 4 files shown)


eumagga0x2a:
An amazing improvement, visually at least, thank you very much! Much closer to what feels "right" with my hardware (which is how it looks when samples are played by mpv with **--vo=gpu**). I do see some green tint with non-clipping RGB adaptive modes for the "The World in HDR" 10-bit AV1 sample. With all samples, the clipping mode is the one which produces the best – really perfect looking – results, except in areas where actual clipping occurs. With other RGB adaptive modes, I need to apply a lot of boost (at least 1.8 with Reinhard, 2 with Hable), even after reducing target luminance to values below 40.

One big request: could you please replace all the magic numbers with maintainable alternatives or at least add comments (probably a lot of comments) explaining the values used?

Once merged, do we still need the previous, inferior options?

szlldm:
> With all samples, the clipping mode is the one which produces the best – really perfect looking – results, except in areas where actual clipping occurs. 

This is a known fact. Clipping is the best to preserves colors, others not. 
`the "hue and saturation shifts" (really mostly saturation shifts) resulting from applying nonlinear curves per channel are an important feature, not a bug.`

Otherwise there is still something wrong with the colors, but have not figured out why.

> With other RGB adaptive modes, I need to apply a lot of boost (at least 1.8 with Reinhard, 2 with Hable), even after reducing target luminance to values below 40.

With my samples those boost values are too much, so there is no good setting, the final adjusments have to be made by the user. The adaptivity really just means, that try to be reasonably good simultaneously at brigth and dark scenes. Target luminance probably has little effect.

> One big request: could you please replace all the magic numbers with maintainable alternatives or at least add comments (probably a lot of comments) explaining the values used?

Those are really magic numbers :'-D

> Once merged, do we still need the previous, inferior options?

Do you mean the non adaptive rgb mappers? There could be use cases, i think...

eumagga0x2a:
> Those are really magic numbers :'-D

This means, the PR provides better working unmaintainable code vs worse working unmaintainable code, thus merged.

> > Once merged, do we still need the previous, inferior options?
> 
> Do you mean the non adaptive rgb mappers? There could be use cases, i think...

Unless the use cases are critically important, I'd advocate keeping the menu options while silently upgrading the functionality.

PR #408, opened 2022-02-16
[hdr] adaptive RGB checkbox + gamut control option

szlldm:
-move adaptive option to a checkbox (default: checked)
-try to match levels between all adaptive RGB tonemappers
-add comments, replace some magic numbers with defines, rename variables to match functionality

### DIFF
--- a/avidemux/common/ADM_commonUI/DIA_HDRConfig.cpp
+++ b/avidemux/common/ADM_commonUI/DIA_HDRConfig.cpp
@@ -11,13 +11,13 @@
 
 #include "DIA_factory.h"
 
-int DIA_getHDRParams( uint32_t * toneMappingMethod, float * saturationAdjust, float * boostAdjust, bool * adaptiveRGB)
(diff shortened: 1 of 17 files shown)


szlldm:
-gamut control option (default: clipping) for RGB mappers (clipping out of gamut is more color accurate)

TODO: it would be nice, if adaptive chkbox & gamut control would be only enabled if a RGB mapper selected. idk how to do it.

eumagga0x2a:
> -add comments, replace some magic numbers with defines, rename variables to match functionality

This helps a lot, thank you very much.

> TODO: it would be nice, if adaptive chkbox & gamut control would be only enabled if a RGB mapper selected. idk how to do it.

Should not be a problem, I'll take care of it, but are you sure we need a higher MENU_MAX_lINK?

Anyway, merging now, refinements can be done later. Thanks!

szlldm:
> are you sure we need a higher MENU_MAX_lINK?

It crashed opening HDR config, complaining about it. Also (5 menu entry) * (4 linked elements) = 20 > 16.

eumagga0x2a:
With adaptive checkbox and gamut control enabled only if a RGB tonemapper selected, we consume 14 elements of the array, which technically fits into the old limit of 16.

Not worth going back, 16 was quite tight anyway.

PR #409, opened 2022-02-18
[hdr] RGB soft limit mapper

szlldm:
this might be the best


eumagga0x2a:
> this might be the best

I'm sorry, but judging from available HDR samples, it is not. At least on my equipment, RGB clipping with compression wins by a large margin.

We might need a pref to make selected out-of-gamut handling other than clipping stick or solve it in a different way if it is relevant for RGB clipping only and that's it, IMHO.

szlldm:
> I'm sorry, but judging from available HDR samples, it is not.

Exactly this is why, there are multiple options :)

>  if it is relevant for RGB clipping only

Gamut clipping/compression is independent from tonemapping (just has to be an RGB mapper)

eumagga0x2a:
Thanks!

PR #410, opened 2022-02-20
[GUI/audioTracks] show up to 32 tracks

szlldm:
-show minimum 4 tracks
-show one inactive track (if possible); if activated, extend the list
-button to duplicate codec and filter config of the first track to other tracks

to maintain the dialog size i used
`this->resize(100,100);  // hacky way to set size to minimum-fit`
which works on linux, but don't now if ok elsewhere...


eumagga0x2a:
Great, thank you very much. All further improvements and polish can happen in-tree, I think.

PR #411, opened 2022-02-22
[prefs] *nix encoding priority

szlldm:
-unused "Indexing priority" removed (only from the Preference dialog)
-new encoding priority for *nixes (only lowering or normal)
 lowering works without privileges

PS: it would be reasonable to remove Playback priority and priority options above normal (imho it is placebo); only encoding priority reduction makes sense (in this case PRIORITY_ENCODING_NIX could have been reunified with PRIORITY_ENCODING)


avidemux/common/ADM_commonUI/DIA_prefs.cpp
```diff
@@ -354,7 +360,20 @@ std::string currentSdlDriver=getSdlDriverName();
...
-
+#else
+        diaMenuEntry priorityEntries[] = {
+                     {2,      QT_TRANSLATE_NOOP("adm","Normal"),NULL}
+                    ,{3,      QT_TRANSLATE_NOOP("adm","Below normal"),NULL}
+                    ,{4,      QT_TRANSLATE_NOOP("adm","Low"),NULL}
+        };
+        diaElemMenu menuEncodePriority(&encodeNixPriority, QT_TRANSLATE_NOOP("adm","_Encoding priority:"), NB_ITEMS(priorityEntries), priorityEntries);
+        diaElemReadOnlyText encodePriorityText(NULL,
+                QT_TRANSLATE_NOOP("adm","After an encoding is finished, the restoration of the original priority is only possible with the proper privileges."),NULL);
```
eumagga0x2a, line 371: I would prefer a very clear hint that this is practically a one-way street affecting everything if I am not mistaken, not just encoding. If you agree, I'll change the text accordingly when merged.

eumagga0x2a:
> it would be reasonable to remove Playback priority and priority options above normal (imho it is placebo)

It might work™ on Windows.

avidemux/common/ADM_commonUI/DIA_prefs.cpp
szlldm: Ok.

szlldm:
i have simpified the DIA_prefs code as well

avidemux_core/ADM_coreUI/src/DIA_encoding.cpp
```diff
@@ -36,8 +36,9 @@ DIA_encodingBase::DIA_encodingBase(uint64_t duration)
 DIA_encodingBase::~DIA_encodingBase( )
 {
     ADM_info("DiaEncodingBase: Destroying\n");
-#ifdef _WIN32
-	setpriority(PRIO_PROCESS, 0, _originalPriority);
+#ifndef __HAIKU__
+    // try to restore, at most nothing will happen
```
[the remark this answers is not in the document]
szlldm: return value check is enough, errno is redundant in this case

eumagga0x2a:
Thanks!

PR #414, opened 2022-03-15
[audiofilter] fix dolby + stereo headphone option

szlldm:


avidemux_core/ADM_coreUtils/src/prefs2.conf
```diff
@@ -97,7 +97,7 @@ uint32_t:playback,                     0,      0,      4
 Default{
 uint32_t:postproc_type,                0,      0,      7
 uint32_t:postproc_value,               0,      0,      5
-uint32_t:downmixing,                   2,      0,      3
+uint32_t:downmixing,                   3,      0,      4
```
eumagga0x2a, line 100: Just FYI, I am going to change the default to Stereo.

Anyway, inserting the headphones option means that all existing Avidemux installations which were using the default Dolby Pro Logic II will silently switch to the older Dolby mode (not that anyone is going to notice that ;-)).

eumagga0x2a (commented):
Thank you very much, in absence of any means to verify the result, it sounds and looks good for me.

eumagga0x2a:
Thanks!

(I actually meant just the user-visible string which I wanted to be in plural, not the enum value name. Of course, the new enum value name is fine too.)

PR #417, opened 2022-03-24
Audiofilter upgrade

szlldm:
-configurable compressor
-equalizer
-channel gain adjusment
-channel delay adjusment
-channel remapping

PS: probably there is a bug in DIA_factory, as DIA_audioFilter only works for now by adding dirtyHack1 & 2 dummy elements (otherwise the first checbox unclickable)


szlldm:
Looks like diaElemFloatResettable is the culprit.

avidemux_plugins/ADM_scriptEngines/tinyPy/src/binding/adm.admPyClass
```diff
@@ -31,7 +31,7 @@
 /* METHOD */ int  pySetFilm2Pal:audioSetFilm2Pal(int,int)
 /* METHOD */ int  pySetNormalize:audioSetNormalize(int,int,int)
 /* METHOD */ int  pySetNormalize2:audioSetNormalize2(int,int,int,int)
-/* METHOD */ int  pySetDrc:audioSetDrc(int,int)
+/* METHOD */ int  pySetDrc:audioSetDrc(int,int,int,float,float,float,float,float)
```
eumagga0x2a, line 34: Rather than startle users who try to run old project scripts with a tinyPy exception, maybe it is worth preserving audioSetDrc signature and turning the function into a wrapper for audioSetDrc2 (or audioSetDrcEx or whatever != old name) with all the additional args?

szlldm: [66735c5](https://github.com/mean00/avidemux2/pull/417/commits/66735c5403bcbdbaf4807da47a5cce4af3004288)

eumagga0x2a:
Thank you very much, I'm going to tweak usage notes in the audio filter dialog a bit prior to refreshing translation files.

PR #419, opened 2022-04-15
[UI] QSettings option for opening filter dialogs maximized

szlldm:
View menu / Filters / Open filter dialogs maximized
If enabled, it can spare the manual resizing of filter's dialogs _every time_.


eumagga0x2a:
I assume that the usability problem this PR addresses is essentially fixed by 41cb9e6343009a36e18128d182a756de9db1130d, isn't it?

szlldm:
Not fixed, but defininetly better. Since there is no "Suspended" state for pull requests, i close this for now...

PR #424, opened 2022-05-22
Filter preferences

szlldm:
Filter specific preferences can be set through filter's UI.
Default resizer algorithm can be changed in resize, fitToSize & zoom
Default padding can be changed in fitToSize & zoom
Rubber band visibility can be changed in preferences of zoom, crop & blackenBorder


eumagga0x2a:
Obsoleted by #425

PR #428, opened 2022-06-09
[GUI] add "Go to Marker" and missing 1/2/4 sec jump buttons + configurable 1min jump buttons

szlldm:


szlldm:
The one minute jump buttons on the navigation widget can be looped through the 1/2/4/60 sec jump options by simply scrolling over the button(s).

eumagga0x2a:
Awesome, thanks.

We should probably either reset the translation table on close or make it persistent via QSettings.

PR #436, opened 2022-06-27
[rotate] option for arbitrary angle rotation

szlldm:


szlldm:
This is undoubtedly the extension of the existing Rotate filter's functionality (the arbitrary rotator would generate the exact same output for 0/90/180/270 degree rotation), i dont see any reason it should be in a different filter.
On-the -fly preview is not possible, it changes dimension (exactly the same way as the 90/180/270 rotation).

> Especially as in the future, such a new filter would allow to integrate further useful functionality like an option to keep width and height of the input and downscale the rotated picture or to select the center of rotation.
This can be already done with the Quadrilateral filter and some math :-) .

eumagga0x2a:
> i dont see any reason it should be in a different filter.

The simple rotation and the arbitrary angle rotation functionality serve distinctly different purposes: the former primarily hardcodes orientation metadata into video, the latter corrects mostly minor and difficult to quantify tilts. The former doesn't need live preview as the only thing to check is 90° vs 270° error. The latter often requires a tedious search for the right angle, which is a pain when it requires closing the configuration dialog and opening the preview in the filter manager after each change (I understand that this is exactly the UX with existing glRotate filter).

> On-the -fly preview is not possible, it changes dimension

Integrating fitToSize functionality purely for preview would allow this, probably intolerably worsening performance when done on the CPU.

szlldm:
IMHO a filter called "Rotate" should do one task, and it is rotate. This PR just extends the former limitations of it.
ACK on inconvenience of the preview.

PR #437, opened 2022-07-07
[GUI] enable autorepeat on 1/2/4/60 timejump buttons

szlldm:


eumagga0x2a:
I don't think this is a good idea, feels like torture for the application and hardware. IMHO, whoever needs quick coarse seek, should drag the navigation slider instead of being enticed to emit a barrage of precise, computationally very expensive seeks. I understand that some other buttons allow autorepeat. These two shouldn't, IMHO.

PR #439, opened 2022-07-12
[GUI] new save button

szlldm:


eumagga0x2a:
Even if not final, this icon immediately removes the moment of hesitation from the brain parsing the arrow direction on otherwise identical folder icons. Thank you very much.

PR #441, opened 2022-07-17
[GUI] add HDR tone mapping button to the Toolbar

szlldm:
easier access when trying out many settings


eumagga0x2a:
Thank you very much, but could you please open a new pull request just for the icon and the resource collection files, either for the original version or for the original version with sliders citing the blue color of other buttons (not sure that this looks good, just an idea) to better differentiate from the "play filtered" button in disabled state? I would implement all the integration myself.

PR #443, opened 2022-07-22
[Q_gui2] Show HDR config button on Toolbar if HDR content loaded

szlldm:


eumagga0x2a:
Thanks, I'll just add the separator stuff.

PR #444, opened 2022-07-23
[FlyDialog] add button for goto the start of the selection

szlldm:


eumagga0x2a:
Thank you very much, the button needs some love, but this can be done in-tree together with necessary bugfixes.

PR #446, opened 2022-07-24
[stillimage] bugfixes

szlldm:


eumagga0x2a:
I believe that [fix goto issues](https://github.com/mean00/avidemux2/pull/446/commits/a7c1016950c25c94d7851ef87aa9d776d84ec0d9) is not right, but the other one is point on. Thanks a lot!

PR #447, opened 2022-07-24
fix audio stretch peak RAM usage; return to old method in case of conversion between Film and PAL

szlldm:


eumagga0x2a:
Thank you very much.

PR #448, opened 2022-07-25
[stillimage] fix stillPicture capture regression

szlldm:


eumagga0x2a:
Thank you, I'm sorry for the extreme slow-motion mid-air collision, keeping my patch in this case.

PR #450, opened 2022-08-03
[ADM_audioStretch] replace rubberband with soundtouch

szlldm:


eumagga0x2a:
Thank you very much, I'll apply the fix for build with VC++ on top of that.

PR #457, opened 2022-10-18
[ADM_mwNavSlider] use lighter blue for marker selection in dark themes

szlldm:


eumagga0x2a:
Thanks, what do you think about using in dark mode [dodgerblue](https://www.w3.org/TR/SVG11/types.html#ColorKeywords) instead, which is even lighter and matches exactly the blue tone used by Qt for e.g. QSlider on macOS?

Else the both blue colors look IMHO too similar while being uncomfortably different (openly different = good, identical = good, very close but not identical = bad).

Just in case, dodgerblue doesn't look good with the light theme, so we cannot use a single blue color in all circumstances, unfortunately.

Whatever color we use, we probably need to change ADM_flyNavSlider accordingly. I wonder whether the color stuff could live in a parent class to avoid code duplication.

PR #459, opened 2022-10-22
[T_QSlider] fix uninitialized variables

szlldm:


eumagga0x2a:
Could have been just `setInvertedWheel(false);`, couldn't it?

I have no idea how I could miss that, thank you for the fix.

PR #461, opened 2022-10-30
[videoFilter] New filter: A.I. Enhance (sw)

szlldm:
depends on #460 
the SSE2 acceleration utilize intrinsic functions, therefore buildability test required in case of clang or MSVC (btw. it should work)
ARM NEON acceleration might be possible via [https://github.com/DLTcollab/sse2neon](https://github.com/DLTcollab/sse2neon)


eumagga0x2a:
Thank you very much. Merging is the only easy way to figure out whether build with VC++ fails, so giving it a try.

PR #464, opened 2022-11-15
fix issue caused by different output stride

szlldm:


eumagga0x2a:
Thank you very much for all the fixes.

PR #465, opened 2022-11-21
[x264] make high10 functional

szlldm:
-bump minimum version of x264 to 153 (back from 2018)


eumagga0x2a:
Thanks!

I hope users understand that by choosing high10 profile they ask for output incompatible with most hw accelerators = unusable on mobile devices.

PR #466, opened 2022-12-12
[GUI] enable keyframe navigation during play;

szlldm:
keep media player style Left/Right button navigation (~4 sec jump)


eumagga0x2a:
Thank you,

+r with the problem of convulsions from brief intermittent resume of playback solved. This issue affects navigation using keyboard shortcuts as well (e.g. no resume of playback until keys or buttons released).

I'll take care of updating tooltips later myself, thanks. They should truthfully reflect the change of the function of the next and previous keyframe buttons during playback.

PR #467, opened 2022-12-24
[videoFilter] New filter: Fields Align

szlldm:
Swap fields, vertical & horizontal adjust


eumagga0x2a:
Thank you.

(Not sure about use case, but the effect is impressive.)

szlldm:
Use case could be to correct analog recordings if there is small error in alignment(s).

PR #468, opened 2022-12-28
CMake: Ensure git rev SHAs aren't abbreviated

ferdnyc:
When building from the repo, `ADM_SUBVERSION` is partly generated from the first 11 characters of a `git log --format=oneline` output line. This can misparse the revision string if they're allowed to be abbreviated, which can leave them at only 8 or 9 characters. The resulting `ADM_SUBVERSION`, e.g. `221228_abcde0123 (`, will break `printf` calls in the code due to the unmatched open paren.

Adding `--no-abbrev` to the command ensures that the full 40-character SHA  is output, which can then be truncated to whatever length the script needs.


ferdnyc:
> The resulting `ADM_SUBVERSION`, e.g. `221228_abcde0123 (`, will break `printf` calls in the code due to the unmatched open paren.

Actually, it might be breaking the `MKSTRING()` macro used _in_ the `printf()` — point is, it breaks something, and it's not supposed to be there anyway.

eumagga0x2a:
Thank you.

PR #470, opened 2023-01-01
[resampleFps/motin] & [imageStab/motest] use SIMD intrinsics

szlldm:
depends on #469


eumagga0x2a:
Thank you. Build with MSVC pending, other platforms seem to be fine on a local test.

PR #471, opened 2023-01-02
[bootstrap] Harmonize builddir names

ferdnyc:
The bootstrap scripts were using "buildCli" as the name of the CLI builddir, but then the CLI plugins were built in a directory named "buildPluginsCLI", a difference in casing that just seems unnecessarily confusing.

I normalized them to "buildCli" and "buildPluginsCli", though I'd be happy to go the other way ("CLI" for both) if preferred.

### DIFF
--- a/bootStrap.bash
+++ b/bootStrap.bash
@@ -325,7 +325,7 @@ fi
 if [ "x$do_plugins" = "x1" -a "x$do_cli" = "x1" ] ; then 
         echo "** Plugins CLI **"
         cd $TOP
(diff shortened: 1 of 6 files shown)


ferdnyc:
(Oh, yes, and as further precedent, there was already this:)

https://github.com/mean00/avidemux2/blob/6063658223e843e4ee559dac090ea8b1bd20923f/bootStrapCrossMingwQt5_mxe.sh#L407-L410

PR #472, opened 2023-01-03
[cmake] Don't set CMAKE_EDIT_COMMAND

ferdnyc:
`CMAKE_EDIT_COMMAND` is mean to be an internal variable, used only for implementing the `make edit_cache` target of Makefile Generators. It's supposed to contain the path to a CMake editor like `ccmake` or `cmake-gui`, and will be called with arguments not accepted by `vim`, `mcedit`, or any other generic editor.

Ref: https://cmake.org/cmake/help/latest/variable/CMAKE_EDIT_COMMAND.html

Here's what happens if you run `make edit_cache` with the `CMAKE_EDIT_COMMAND` set to `vim` as in the bootstrap scripts:

```console
$ cd buildCli
$ make edit_cache
Running CMake cache editor...
vim -S.../avidemux2/avidemux/cli -B.../avidemux2/buildCli
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Dec 14 2022 00:00:00)
Garbage after option argument: "-S.../avidemux2/avidemux/cli"
More info with: "vim -h"
make: *** [Makefile:74: edit_cache] Error 1
```

### DIFF
--- a/bootStrap.bash
+++ b/bootStrap.bash
@@ -65,7 +65,7 @@ Process()
                 mkdir $BUILDDIR || fail mkdir
         fi
         cd $BUILDDIR 
-        cmake $COMPILER $PKG $FAKEROOT $QT_FLAVOR -DCMAKE_EDIT_COMMAND=vim $INSTALL_PREFIX $EXTRA $BUILD_QUIRKS $ASAN $DEBUG -G "$BUILDER" $SOURCEDIR || fail cmakeZ
+        cmake $COMPILER $PKG $FAKEROOT $QT_FLAVOR $INSTALL_PREFIX $EXTRA $BUILD_QUIRKS $ASAN $DEBUG -G "$BUILDER" $SOURCEDIR || fail cmakeZ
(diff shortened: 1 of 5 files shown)


eumagga0x2a:
Thank you.

Haiku and macOS Sierra and High Sierra scripts are entirely unused legacy files, kept not necessarily for objective reasons. Correcting code there is definitely a waste of effort.

ferdnyc:
@eumagga0x2a 

> Haiku and macOS Sierra and High Sierra scripts are entirely unused legacy files, kept not necessarily for objective reasons. Correcting code there is definitely a waste of effort.

That's good to know, thanks, and I appreciate your consideration of my time/effort... but I probably would've done it regardless. It was a matter of a few seconds' work, almost would've taken more effort to be surgical than just fix everything. Plus, while the scripts _are_ there, leaving them "out of sync" would've just felt a bit wrong.

But I'll definitely avoid any **major** changes involving those outdated tools, should I consider anything like that.

PR #473, opened 2023-01-04
[aiEnhance] fix typo

szlldm:


eumagga0x2a:
At the right moment before the typo could make it into translation files. Thanks!

PR #474, opened 2023-01-10
[aiEnhance] store network parameters as int16; add fast x2 upscaler with denoising

szlldm:
storing degraded parameters as int16 reduces plugin size, without affecting output quality


eumagga0x2a:
Thank you very much. Build tests still pending.

PR #477, opened 2023-02-07
[GUI_xvRender] initialize to black

szlldm:
this prevent green flash on video loading


szlldm:
VDPAU renderer should also get a similar treatment, it flash garbage on load

eumagga0x2a:
Thanks!

PR #487, opened 2023-03-24
[GUI] use less confusing cursor to indicate busy state

szlldm:


ferdnyc (approved):
This probably makes sense, IMHO. Qt's [cursor definitions](https://doc.qt.io/qt-6/qt.html#CursorShape-enum) for the two are:

> `Qt::WaitCursor` - An hourglass or watch cursor, usually shown during operations that prevent the user from interacting with the application.
> `Qt::BusyCursor` - An hourglass or watch cursor, usually shown during operations that allow the user to interact with the application while they are performed in the background.

The example images show the WaitCursor as an hourglass, while the BusyCursor is a pointer with a small hourglass attached. (Which is how they'll look on Windows. Both Linux/GNOME and macOS use a spinner/beachball animation for the WaitCursor, and IIRC both use a pointer with attached spinner/beachball for the BusyCursor. I know Linux does, at least.)

The user is not prevented from interacting with Avidemux during transport actions (seeking, skipping, etc.). In fact, clicks can be repeated freely even while "busy", and held down to autorepeat. This makes `WaitCursor` a misleading choice.

eumagga0x2a:
Thanks!

PR #489, opened 2023-04-05
Statusbar continuance

szlldm:


szlldm:
>  I guess, we don't need to yell at the user anymore as the context (error / warning/ info) is now clear thanks to the statusbar icon?

Right, it's a valid point. I would keep the ERROR capitalised though.

eumagga0x2a:
I'm about to remove them alltogether, relying on the the icon. Do you see a sufficient reason to duplicate it in the message text?

szlldm:
>  Do you see a sufficient reason to duplicate it in the message text?

I have a feeling, the icon is not enough on its own.
The message box also has the icon & the "duplicate" text

eumagga0x2a:
> The message box also has the icon & the "duplicate" text

I feel that hardcoding "Info" as window title and the resulting necessity to counteract this by adding a more specific "title" for `QMessageBox::critical` in the Qt implementation of `GUI_Error_HIG()` is rather a bug than a template worth to imitate. Not really important, however.

PR #491, opened 2023-04-25
[T_thumbSlider] fix issue "doesn't get the focus on the first mouse click"

szlldm:


eumagga0x2a:
I see, there was no need to ignore mouse press event for thumb slider, introduced by cd206a7, on failure to seek to previous or next keyframe. Same should apply to mouse move event.

Thanks!

szlldm:
> Same should apply to mouse move event.

This would worsen the UX: movable wheel without any navigation after a stop.

PR #494, opened 2023-06-01
[videoFilter] New filter: Grain

szlldm:


eumagga0x2a:
Thank you very much. It would be great if at least some of the magic numbers could be replaced by defines or their meaning clarified in comments though, maybe in a follow-up PR, please?

PR #495, opened 2023-06-06
[tinyPy] fix autogenerate

szlldm:
- fix generator (produces exactly the same files from the same bindings)
- fix binding
- regenerate all


szlldm:
"fix autogenerate script" should have been "fix autogenerate makefile"

eumagga0x2a:
Thank you very much.

PR #496, opened 2023-06-10
Scripting shell & TinyPy improvements

szlldm:


eumagga0x2a:
Great, thanks. os.help() throws an error, but this can be handled in a follow-up.

PR #500, opened 2023-08-02
Fix missing symbols on PowerPC

barracuda156:
Looks like some current defines are wrong, and they break the build.
Fix that.


barracuda156:
P. S. There is one more place where existing code is wrong, which is not addressed here: https://github.com/mean00/avidemux2/blob/b564488a9c36d42290decc5a1671eff96e178ab7/cmake/admDetermineSystem.cmake#L54-L99
Locally I simply set PowerPC as a fallback value – instead of erring out (well, because locally I know for sure).
There should be an appropriate check akin to other archs. (In fact CMake has all needed defines, I am not sure what was the need to use such an extravagant method, but w/e.)

eumagga0x2a:
Thank you. I must admit that PPC is completely off my radar.

barracuda156:
@eumagga0x2a Thank you for merging!
I did not yet try the app in action, but now it builds on PPC, launches fine and menus work.

PR #503, opened 2023-10-24
[ADM_audio] s32le & s64le PCM

szlldm:


eumagga0x2a:
Thank you, I'll take care of the Mp4 demuxer myself.

Not sure about pcm_s64le, it doesn't seem to be a trivial way to detect it in MOV files. Do we really need to support it?

szlldm:
I thought MP4 does not support PCM.

PR #507, opened 2023-12-03
Proposal: improve organization

plata:
PR because discussions and issues are not available.


ofaurax:
Thanks for opening this. This has been 30 min since I search when I can post a crash report about avidemux.....

PR #509, opened 2023-12-09
[Q_gui2] fix "Light theme" palette; always show "Default theme" option

szlldm:


eumagga0x2a:
Thank you very much for this fix.

PR #512, opened 2024-03-08
[GUI] User scripts on the Toolbar

szlldm:


eumagga0x2a:
Thank you for quickly taking the initiative, would you please mind addressing the issue of the script toolbar buttons not being re-enabled after stopping playback?

Not liking the icons (would prefer re-using the Properties icon as base especially without introducing a new color not reflected in any other part of the application window), but this doesn't matter for now. Would block the release on this, however.

eumagga0x2a:
Thank you very much, please open new PRs for further refinements, if necessary.

PR #513, opened 2024-03-10
[GUI] new Toolbar script buttons

szlldm:


eumagga0x2a:
Looks great, thank you very much.

PR #514, opened 2024-04-07
[fitToSize] swap dimensions button

szlldm:
using UTF-8 "⇄" as button text
hopes it works anywhere


eumagga0x2a:
Thank you, I'll address two minor wishes myself.

PR #516, opened 2024-04-26
[MpegTS] index file in memory 

szlldm:
if cannot create idx file


eumagga0x2a:
Looks good, thank you very much.

PR #517, opened 2024-04-27
[Matroska] index to file option

szlldm:


eumagga0x2a:
Thank you very much, I've changed some aspects which bothered me and enabled the feature by default to get some testing coverage.

PR #518, opened 2024-07-13
[ADM_videoEncoder] VA-API AV1 encoder

szlldm:
Untested


eumagga0x2a:
Briefly tested working, will do cleanup similar to one applied to H.264 and HEVC VA-API encoders myself in-tree. Thank you very much.

PR #519, opened 2024-07-18
Add file system case-sensitive check

denizt:
The build requires the file system to be case-sensitive and will proceed and fail even if it's case-insensitive (which is the default on macOS, unfortunately).

Add isCaseSensitive function to avoid this pitfall early on.

This PR also tries to improve documentation by highlighting this requirement and removing the extra step for git submodules. After finalising the review/feedback and the script, I want to add this to other Mac build scripts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Avidemux is a simple cross-platform video editor for Linux, Windows and macOS.
 # Build from source
 
 To get Avidemux source code from the main repository and the translations,
-run the following commands in a directory located on a case-sensitive file system:
+run the following commands in a directory located on a **_case-sensitive_** file system:
 ```
 git clone --recursive https://github.com/mean00/avidemux2.git
 cd avidemux2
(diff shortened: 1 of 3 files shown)


eumagga0x2a:
I'm sorry for invalidating this PR with the work on out-of-tree build. Bailing out on case-insensitivity is a good idea. If you agree to open a new pull request, the implementation must be compatible with out-of-tree build and should not preclude the bootstrap script from printing usage instructions.

Thank you.

denizt:
Would it be possible for you to take the function and use it as you wish? I am rather swamped right now, and I would appreciate it if the build had this check and saved everyone the time (and the sanity 😄) especially considering there is no Mx Arm build for macOS.

eumagga0x2a:
[Done](https://github.com/mean00/avidemux2/commit/38582543d2933b780e72df9f3a0b6bddf057a91a), thank you again.

PR #520, opened 2024-07-25
New index to file helper class; Applied on MKV

szlldm:


eumagga0x2a:
Awesome, thank you very much.

PR #521, opened 2024-08-16
[ADM_demuxers/Mp4] index to file option

szlldm:
only saves frame flags, to skip the long "Decoding frame type" process
if "Decoding frame type" not triggered, then index is not created, bc. the file already opens quickly


eumagga0x2a:
Thank you very much.

PR #531, opened 2024-10-14
[delogoHQ] fix rare division by zero cases

szlldm:


eumagga0x2a:
Seems to be not that rare, by all means needed, thanks.

PR #532, opened 2024-11-04
Add support for loongarch64

YHStar:
[LoongArch](https://docs.kernel.org/arch/loongarch/introduction.html) is a new RISC ISA developed by Loongson. There are already a lot of [community support and testing](https://www.phoronix.com/search/LoongArch) about it.

Hello, I'm working on porting ArchLinux to Loongarch64. Would you like to review my changes and support Loongarch64? It's tested on LoongArchLinux.


mean00:
Hi Thank you for that
The main concern is that we dont have the hw to check an upcoming patch is not breaking the loongarch64 support
(which looks a lot like rv64 btw).
Do you know if there is a version of Qemu supporting linux on top of loonggarch64 to do a minimum testing  ?
We can also take the patch, it is not intrusive and just classify loongarch64 in the unsupported section

YHStar:
Hello, thank you for your review! 
Loongarch has been supported by QEMU *officially* since 7.2.0, so just use QEMU Official latest you can get a loongarch environment.

Here are some loongarch LiveCDs I've tested :
> - [Debian unstable](https://cdimage.debian.org/cdimage/ports/tests/loong64-test-20241115-2/) 
> - [LoongArchLinux by LCPU](https://github.com/lcpu-club/loongarch-packages/releases/tag/2024.12)

Here are the edk2 for loongarch
> - [ekd2-platforms - tianocore](https://github.com/tianocore/edk2-platforms)

I have tested the package on my loongarch machine, and if you have any problem just tell me.  :)

PR #534, opened 2024-12-01
[GUI_sdlRender] fix HiDPI, fix preference saving

szlldm:


avidemux/common/ADM_render/GUI_sdlRender.cpp
```diff
@@ -51,7 +51,9 @@ static std::vector<sdlDriverInfo>listOfSDLDrivers;
 class sdlRenderImpl: public VideoRenderBase
 {
   protected:
+              GUI_WindowInfo info;
```
eumagga0x2a, line 54: I guess we don't need the full struct, just a double for scalingFactor, do we?

avidemux/common/ADM_render/GUI_sdlRender.cpp
```diff
@@ -288,7 +299,10 @@ bool sdlRenderImpl::init( GUI_WindowInfo *window, uint32_t w, uint32_t h, float
 
     int bpp;
     int flags;
+    info=*window;
     baseInit(w,h,zoom);
+    displayWidth*=info.scalingFactor;
+    displayHeight*=info.scalingFactor;
```
eumagga0x2a, line 305: Maybe factorized like in libvaRender?

eumagga0x2a:
I can address the nits post-merge, if you agree.

avidemux/common/ADM_render/GUI_sdlRender.cpp
szlldm: maybe in the future other members will be needed. there is literally no gain if we only store the scalingFactor

PR #536, opened 2025-01-13
[GUI_xvRender] allow multiple instance to use Xv render

szlldm:
unbreaking PR #527


eumagga0x2a:
Thank you.

Doesn't seem to do any harm, concurrent execution remaining unsupported (but to some extent working).

PR #542, opened 2025-03-16
Fix install path for Debian

marillat:
Without this patch files are installed in

`/lib` instead of `/usr/lib`


eumagga0x2a:
Thank you, there are wrong assumptions about directory structure in our code, to be fixed.

PR #544, opened 2025-03-28
Exec options must be %U instead of %f

marillat:
From Desktop Entry Specification: https://specifications.freedesktop.org/desktop-entry-spec/1.1/exec-variables.html

%U is preferred to open multiples file or URLs

As example in an Debian unstable 
```
$ grep Exec /usr/share/applications/*|grep -c %f
14
marillat~  
$ grep Exec /usr/share/applications/*|grep -c %U
35
```

### DIFF
--- a/avidemux/qt4/xdg_data/org.avidemux.Avidemux.desktop.in
+++ b/avidemux/qt4/xdg_data/org.avidemux.Avidemux.desktop.in
@@ -2,7 +2,7 @@
 Name=Avidemux
 GenericName=Video Editor
 Comment=Multiplatform video editor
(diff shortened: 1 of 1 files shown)


eumagga0x2a:
Thank you. Currently neither Avidemux is capable of handling multiple URLs in the way `%U` is interpreted by a desktop environment nor multiple instances of Avidemux should be launched simultaneously. A possible solution would be treating each URL list entry after the first one as an argument for the `--append` option (UX with incompatible files may be pretty bad then). With that done, changes to the desktop entry may be considered.

marillat:
OK, I'm closing this PR

PR #545, opened 2025-04-07
[Q_gui2] fix application icon regression

szlldm:


eumagga0x2a:
Well noted, thanks! However, this will be broken the very next time clang-format tidies up the file. I had no idea that choosing an alias with a hyphen instead of an underscore in 5980785 may be dangerous in any regard. I'll add special comments as protection.

PR #550, opened 2026-04-30
Update README.md

messimuc:
couldnt compile without relaxed checks on macOS Tahoe


eumagga0x2a:
Please provide more details. Do you have any other CXXFLAGS in the env? Build succeeds on my Tahoe 26.4.1 system without these options. (However, packaging turned out to be broken, now fixed).

PR #552, opened 2026-06-21
demuxers/Matroska: add FFV1 codec ID mapping

ShortyCM:
Maps Matroska CodecID V_FFV1 to internal FourCC FFV1.

Without this mapping, FFV1-in-MKV opens with Video FCC 00000000, decoder lookup fails, and Avidemux falls back to the RGB/raw codec.

Tested with an FFV1 v3.4 MKV, 720x480, YUV 4:2:0 8-bit. After this one-line mapping, the file opens using the lavcodec decoder path and displays correctly.


mean00:
Thanks!

PR #555, opened 2026-08-20
Fix native arm64 (Apple Silicon) Homebrew build on macOS

rogerawong:
Native arm64 builds on Apple Silicon via `bootStrapMacOS_Monterey.arm64.sh` with Homebrew fail at two points. Both are path assumptions from the Intel/`/usr/local` era.

**1. SDL3 link failure (`qt4`)**

```
ld: library 'SDL3' not found
clang++: error: linker command failed with exit code 1
make[2]: *** [ADM_render_qt/libADM_render6_QT6.dylib] Error 1
```

`ADM_render_qt` links `${SDL3_LIBRARIES}` (the bare `SDL3` name) and the main app links a literal `SDL3`, so the `-L` directory reported by `pkg-config --libs sdl3` is never used. That works on Linux where `/usr/lib` is a default linker search path, but fails with Homebrew on Apple Silicon where `/opt/homebrew/lib` is not. Fixed by linking `${SDL3_LINK_LIBRARIES}` (absolute paths, populated by `pkg_check_modules` alongside `SDL3_LIBRARIES`) at both sites. No behavior change on platforms where the bare name already resolved.

**2. `fixup_bundle` failure during packaging (`osxInstaller`)**

```
CMake Error at BundleUtilities.cmake:725 (message):
  otool -l failed: 1
  error: otool-classic: can't open file:
  @rpath/QtCore.framework/Versions/A/QtCore (No such file or directory)
CPack Error: Error when generating package: Avidemux
```

`bundleDirs` hardcodes `/usr/local/lib`, the Intel Homebrew prefix. On Apple Silicon, Homebrew lives in `/opt/homebrew` and `/usr/local/lib` does not exist, so `fixup_bundle` cannot resolve the `@rpath/Qt*.framework` references carried by the Qt platform and style plugins (the app binaries themselves link Qt by absolute path, which is why only the plugins trip this). The fix selects the prefix by `ADM_CPU_ARM64`, the same idiom the VapourSynth relinking in `admFixupBundle.cmake.in` already uses.

**Tested:** macOS 26.5 arm64, Homebrew Qt 6.11.1 / SDL3 3.4.14, CMake 4.4.2. A clean `bash bootStrapMacOS_Monterey.arm64.sh` run with no extra environment variables completes through DMG creation and ad-hoc signing. Every Mach-O in the resulting bundle is arm64-only, the app launches, and a copy/copy cut of an H.264/AAC MP4 saves and decodes clean under `ffmpeg -v error`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


eumagga0x2a:
Thank you, please clarify why the patch for packaging hack, probably fully obsolete by now, should be necessary. Packaging works for me without it on my fully updated Tahoe 26.6.2.

The SDL3 linking fix is however really needed, would merge it isolated from the other one, thank you again.

edit: SDL3 doesn't work in Avidemux on macOS the way it is used. If a fix cannot be found soon, I'll disable it.

rogerawong:
Thank you! Dropping the packaging change. It turns out the build where I hit that failure had an early LIBRARY_PATH workaround for the SDL3 link error. The SDL3 linking fix makes that workaround completely unnecessary and it packages fine. BUT…

If anyone else has /opt/homebrew/lib in the LIBRARY_PATH, like you might if you're troubleshooting a Homebrew failure, they're going to see the same failure I did. CMake assumes what's in LIBRARY_PATH is a standard system directory, and doesn't include it in the binary's rpaths. So the packaging then craps out on ARM Macs where avidemux's fallback points to /usr/local/lib and not /opt/homebrew/lib.

So I've made a new version that's hardened against this failure (#556). The build system knows where Qt is, so we can just add its location to the packaging search path: /opt/homebrew/lib on ARM Macs, /usr/local/lib on x86 Macs, or your custom Qt lib folder if you build Qt yourself.

It's a more elegant PR for a more civilized age.

I tested packaging with SDL3 off and then on, and both paths package correctly.

eumagga0x2a:
Thank you very much for the linking fix and for identifying the genesis of the problem with app bundle creation.

PR #557, opened 2026-09-18
ffNvEnc stuff and things

poltsy:
Boring uninteresting stuff, cleaning up the mangled presets and tunes, and defaults.


eumagga0x2a:
Thank you very much. I have no way to check whether the new AV1 encoder works, I assume you have verified that it does. As nv-codec-headers version in official build env remains at sdk/11.1 in order to be compatible with old drivers, the required functionality in libavcodec won't be available there.

